PR #7, opened 2018-08-29
Document env vars required for use with TLS

tombentley:


scholzj (commented):
TBH, I think this is a bit confusing. We do not really expect anyone to use the `TRUSTSTORE_*` and `KEYSTORE_*` options. When you just add the other options here he user will not know what to use. Should we remove the keystore and truststore options?

ppatierno (commented):
I was going to ask the same thing, I think that we should remove the other options then and not giving two different ways for doing the same thing in a hello world example.

ppatierno (approved):

tombentley:
@scholzj are you happy with this now?

scholzj (approved):

PR #13, opened 2019-09-14
Add OAuth support and examples

scholzj:
This PR adds support for OAuth and related example deployment file.

### DIFF
--- a/deployment-oauth.yaml
+++ b/deployment-oauth.yaml
@@ -67,7 +67,7 @@ spec:
     spec:
       containers:
       - name: hello-world-producer
-        image: scholzj/hello-world-producer:latest
+        image: strimzi/hello-world-producer:latest
         env:
           - name: BOOTSTRAP_SERVERS
             value: my-cluster-kafka-bootstrap:9092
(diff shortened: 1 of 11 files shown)


deployment-oauth.yaml
```diff
@@ -0,0 +1,260 @@
+apiVersion: kafka.strimzi.io/v1alpha1
```
ppatierno, line 1: v1beta1?

deployment-oauth.yaml
```diff
@@ -0,0 +1,260 @@
...
+kind: KafkaTopic
+metadata:
+  name: my-topic
+  labels:
+      strimzi.io/cluster: my-cluster
+spec:
+  replicas: 3
+  partitions: 12
+---
+apiVersion: kafka.strimzi.io/v1alpha1
```
ppatierno, line 11: v1beta1?

ppatierno (approved):
LGTM. Just a couple of nits.

deployment-oauth.yaml
scholzj: This is what we have also in all other examples. It has also the big advatnage that it works with older versions when you need to try them because of some questions. So I would keep this.

ppatierno: but we are already using v1beta1 for KafkaUser here

scholzj: I guess I copied it from other place. I changed it to `v1alpha1`. If we want to change them all to `v1beta1` we should probably use different PR as it is much more files than just this one.

ppatierno: Agree

PR #14, opened 2019-09-15
Use apps/v1 instead of extensions/v1beta1

scholzj:
Kubernetes 1.16 drops support for extensions/v1beta1. We should update our Deployments to apps/v1 to make sure we are compatible.


ppatierno (approved):
LGTM

PR #15, opened 2019-10-16
Send by default more than 1000 messages

scholzj:
The example deployment YAML files right now always send/receive only 1.000 messages and afterwards they exit. This sucks when using them for demos, because the processes will start restarting in the middle of a demo. This PR increases the count to 1.000.000 to make sure they last longer. WDYT @ppatierno?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: required
 language: java
 jdk:
-- oraclejdk8
+- openjdk8
 services:
 - docker
 script:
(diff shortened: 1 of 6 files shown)


ppatierno (approved):
I am fine with that but maybe by default we could not set the env var at all with the meaning of sending infinite messages until the process is stopped.

scholzj:
@ppatierno I could change the code if you want. But right now the default is only 10 messages and not infinity :-o.

PR #18, opened 2019-11-04
Added Kafka - HTTP bridge examples

### LINKED ISSUE #17: Provide Kafka HTTP client examples
I would move my current Kafka HTTP client examples under this repo and adding the support for the tracing as well.

ppatierno:
This PR fixes #17


ppatierno:
I am going to add Jaeger tracing with this PR as well, in order to fix #19

http/vertx/kafka-http-consumer/src/main/java/ConsumerApp.java
```diff
@@ -0,0 +1,76 @@
+import java.util.Map;
```
scholzj, line 6: Should we have the ASL headers in the Java files?

http/vertx/scripts/run.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
scholzj: Is this script in any way different from the one we have for the regular Kafka clients? If not, should we have it just once?

scholzj (commented):
I left some comments, but the code looks good.

scholzj (approved):

PR #27, opened 2020-02-26
Fix an NPE when no CLIENT_RACK is specified

ppatierno:
When no CLIENT_RACK is set, an NPE is raised when putting a null value in the Properties bag for consumer configuration.


scholzj (approved):
Ups. Sorry.

PR #32, opened 2020-05-27
Add option to better configuraiton of example clients

Frawless:
This PR adds option to specify more parametrs for example clients. We want to use these clients in strimzi system tests and this change will be very helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,8 @@ Producer
 * `USER_CRT` - the user's certificate
 * `USER_KEY` - the user's private key
 * `LOG_LEVEL` - logging level  
+* `PRODUCER_ACKS` = acknowledgement level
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -63,6 +63,8 @@ Producer
 * `USER_CRT` - the user's certificate
 * `USER_KEY` - the user's private key
 * `LOG_LEVEL` - logging level  
+* `PRODUCER_ACKS` = acknowledgement level
+* `ADDITIONAL_CONFIG` = additional configuration for producer
```
scholzj: You should add some explanation abotu the format - that it is a key=value map etc.

scholzj (approved):

ppatierno (approved):

PR #40, opened 2020-08-27
[MO] - [logging] -> fix broken logging + support for oauth logging

see-quick:
This PR fixing the logging, which did not work for the clients. Moreover adding debug info about the config of each client.

Signed-off-by: morsak <xorsak02@stud.fit.vutbr.cz>

### DIFF
--- a/java/kafka/pom.xml
+++ b/java/kafka/pom.xml
@@ -16,7 +16,7 @@
     <properties>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
-        <log4j.version>2.13.3</log4j.version>
+        <log4j.version>2.7</log4j.version>
         <kafka.version>2.4.0</kafka.version>
         <opentracing-kafka.version>0.1.11</opentracing-kafka.version>
         <jaeger.version>1.1.0</jaeger.version>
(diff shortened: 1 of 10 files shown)


java/kafka/pom.xml
```diff
@@ -16,7 +16,7 @@
     <properties>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
-        <log4j.version>2.13.3</log4j.version>
+        <log4j.version>2.7</log4j.version>
```
scholzj: Why are we downgrading to older version? I would not expect that is something we want to do.

Frawless (approved):
LGTM, but I agree with Jakub about log4j version

java/kafka/pom.xml
see-quick: Hmmm, that was version, which was suggested by Marko but i can try it with the `2.13.3`

see-quick: Fixed

scholzj (approved):

ppatierno (approved):

scholzj:
@see-quick Could you look at the conflict please?

see-quick:
@scholzj Done :)

PR #45, opened 2020-10-29
Renamed hello-world Java to java-kafka

ppatierno:
The current Java examples, using the "native" Kafka Java clients, were born as simple "hello world" applications even just sending "Hello World" text as hardcoded.
Since then they are evolved with more configuration parameters and I would say that still calling them "hello world" is wrong, so this PR suggests to change the prefix in "kafka-native" compared to the "kafka-http" we have for example.
After this PR, if accepted and merged, I will open a new one on the operator repo for changing the Docker images to use.


scholzj:
TBH, I'm not sure about this. The `kafka` part of the name gets duplicated from the directory right above. The `native` part is easily misleading since when coming from the outside it might not be clear what does it mean. Renaming the images is also more mess then help because it would leave the old ones abandoned and confuse the users. So I wonder if this is wroth it and if so, whether this is the right name.

ppatierno:
I think that it's worth doing it maybe choosing a better name. Do you have a proposal? The kafka part in the artifact doesn't have any relation with the above directory even if it's a duplication. Maybe instead of "native" we can use "java" having `kafka-java-consumer`. I am thinking when we'll have more examples in different languages. Of course the directory structure will have the language name but the artifact and the Docker image will have to reflect it, so i.e. `kafka-go-consumer` and so on.
Maybe just `java-consumer`, `go-consumer` and we should also change `java-http-consumer` to not misleading with a future `go-http-consumer` (that I would like to add).
We can live with old images not used anymore, we did in the past as well.
This "hello world" prefix and naming sounds really really bad to me.

scholzj:
> I think that it's worth doing it maybe choosing a better name. Do you have a proposal? The kafka part in the artifact doesn't have any relation with the above directory even if it's a duplication. Maybe instead of "native" we can use "java" having kafka-java-consumer. I am thinking when we'll have more examples in different languages. Of course the directory structure will have the language name but the artifact and the Docker image will have to reflect it, so i.e. kafka-go-consumer and so on.
Maybe just java-consumer, go-consumer and we should also change java-http-consumer to not misleading with a future go-http-consumer (that I would like to add).

I think you should consider the structure which is build to navigate. It already has Java as well as KAfka in it. So why do you want to duplicate it? If you want to call it something else, I would probably follow the APIs they are based on or try to explain that these are the regular plain Kafka clients.

> We can live with old images not used anymore, we did in the past as well.

We did it in the past and it causes a lot of problems from users thinking that they are using the latest versions. 

> This "hello world" prefix and naming sounds really really bad to me.

For me this is exactly what they are - they are just some very simple _Hello world_ applications which show just the basics how to use it. The name does not come from the text they send it comes from the purpose they fulfil in my opinion.

ppatierno:
> I think you should consider the structure which is build to navigate. It already has Java as well as KAfka in it. So why do you want to duplicate it?

When you have the artifact and the Docker image you don't see any source structure to follow.

> If you want to call it something else, I would probably follow the APIs they are based on or try to explain that these are the regular plain Kafka clients.

So what's your proposal here `plain-consumer`? if you don't want `kafka` in the name? Tbh it sounds pretty not understandable.

> We did it in the past and it causes a lot of problems from users thinking that they are using the latest versions.

Actually, these are just client examples so I don't see any big problem

> For me this is exactly what they are - they are just some very simple Hello world applications which show just the basics how to use it. The name does not come from the text they send it comes from the purpose they fulfil in my opinion.

I would agree on the "Hello World" nature and what "Hello World" means to developers but having it in the name is still pretty bad as clients application. Why our HTTP clients are not called "hello-world-http-producer" then?

scholzj:
> I would agree on the "Hello World" nature and what "Hello World" means to developers but having it in the name is still pretty bad as clients application. 

So if you agree it describes well the purpose and functionality, why is it a bad name?

> Why our HTTP clients are not called "hello-world-http-producer" then?

I don't know - I assume you named them.

ppatierno:
> So if you agree it describes well the purpose and functionality, why is it a bad name?

My feeling is that it's bad to reflect explicitly the nature in the name and maybe a real "hello world" would not provide all the customization we provide today on the application. It was a good name at the beginning imho.

> I don't know - I assume you named them.

yeah, because "hello world" was really bad ... this is why this new PR :-)

scholzj:
> My feeling is that it's bad to reflect explicitly the nature in the name

So should we call it `elvghkzdsg` to make sure it does not reflect what it is?

ppatierno:
> So should we call it `elvghkzdsg` to make sure it does not reflect what it is?

No, I was mostly thinking about `hjkhghfghfdtyydsdaft5r5afdgs` ... it sounds better

ppatierno:
@samuel-hawker @Frawless @tombentley we need more opinions for reaching a quorum here :-) ... and "I don't care" is not valid this time ;-)

samuel-hawker:
This is difficult, I'm probably leaning into the `plain-consumer/producer/stream` camp here, I'm not really understanding of what `kafka-native` means in this context, and I agree also that hello-world is not appropriate.
Calling it plain should be okay, or maybe `sample-*` or `basic-*`

ppatierno:
@samuel-hawker yeah I am ok with any other clean suggestion. Maybe what I am proposing as `kafka-native` is not the best one.
Maybe `plain` could be misleading with the "plain" listeners in Kafka so not using TLS.
I like the `sample-*` or `basic-*` idea.
The main point is removing this "hello-world" :-)

tombentley:
I think:

* `native` is too ambiguous to be helpful. 
* the kafka part is pretty much a given for anyone who know what Strimzi is about.
* Call it "hello world" suggests to me that it's where you start when trying to understand this stuff. And from there you move on to more sophisticated examples, which might be in Java or other languages
* If we were to add some golang example it would necessarily to tied to a particular client library, so including that in the name would make more sense than just the language.

So overall I think it's better to stick with the existing names, but not see them as setting a precedent for future naming.

ppatierno:
> Call it "hello world" suggests to me that it's where you start when trying to understand this stuff. And from there you move on to more sophisticated examples, which might be in Java or other languages

@tombentley why not using `basic-*` instead of `hello-world-*` to highlight this point?

> If we were to add some golang example it would necessarily to tied to a particular client library, so including that in the name would make more sense than just the language.

What are you suggesting here? 

> ... but not see them as setting a precedent for future naming

So having examples with a completely different naming from tech to tech, language to language?

tombentley:
> > Call it "hello world" suggests to me that it's where you start when trying to understand this stuff. And from there you move on to more sophisticated examples, which might be in Java or other languages
> 
> @tombentley why not using `basic-*` instead of `hello-world-*` to highlight this point?

That adds little extra (AFAICS), and comes at the cost of renaming something which would be easily missed by someone using the existing images. as Jakub described. 

> 
> > If we were to add some golang example it would necessarily to tied to a particular client library, so including that in the name would make more sense than just the language.
> 
> What are you suggesting here?

Something like sarama-producer

> 
> > ... but not see them as setting a precedent for future naming
> 
> So having examples with a completely different naming from tech to tech, language to language?

I think if we have dozens of examples having a naming schema would be beneficial. But we don't and AFAIK we're not planning on adding lots more, so the downsides of renaming now outweigh the upside of having a naming convention for a handful of things.

ppatierno:
We are moving to quay.io so anyway we are going to suggest people use `quay.io` instead of `docker.io` for images, so changing the name is not a great burden here.

ppatierno:
I had a chat with @scholzj and even because we are moving to `quay.io` it could be the right time to find a good pattern for image names for different examples (so even moving away from using "hello world" :-))
We were thinking about something like:

`<language>-<technology|framework|library>-<producer|consumer|streams>

For example:

* java-kafka-producer ... using the native Java Kafka client
* go-sarama-consumer ... in Go using Sarama as library
* java-vertx-producer ... using the Java Vert.x client
* java-http-vertx-producer .. using HTTP with Vert.x
* go-http-consumer ... in Go using HTTP

Wdyt? @scholzj @tombentley

ppatierno:
As per #46 I renamed the `hello-world` stuff to `java-kafka`.
After this PR, for following #46 I will open another one for moving the "http" folder under the "java" one so following the pattern `java-http-vertx-producer|consumer`.

@scholzj @tombentley @samuel-hawker @Frawless can you review again, please?

scholzj:
Why do you wanna move the HTTP ones under Java folder? I think they are good where they are.

ppatierno:
My idea was to follow the naming convention in the folder structure as well.
On #46 we said that as image we have `java-http-vertx-producer` because in the end HTTP Vert.x examples are still Java.
Then we can have for example `go-sarama-....` and `go-http-.....`. 
The root of all examples should be the language.

scholzj:
I think the fact that the HTTP example do not use Kafka at all is significant enough to keep them separate.

ppatierno:
But the HTTP examples are using Java ...
The tree would be:

* java --> kafka --> consumer|producer|streams
* java --> http --> vertx --> consumer|producer (using Vert.x HTTP client)

and in the future we could have

* java --> vertx --> consumer|producer (using the Vertx Kafka client for example)
* java --> http --> consumer|producer (using same Java available HTTP client?)
* go --> sarama --> consumer|producer
* go --> http --> consumer|producer

etc etc

I don't want to put HTTP under Kafka but under Java.
As I mentioned with the pattern name related to the Docker images we are kind of using the language as root (`<language>-<technology|framework|library>-<producer|consumer|streams>`) so following same way in the folders structure could make more sense.

java/kafka/consumer/Makefile
```diff
@@ -1,4 +1,4 @@
-PROJECT_NAME=hello-world-streams
+PROJECT_NAME=java-kafka-consumer
```
scholzj, line 1: This looks wrong.

java/kafka/consumer/Dockerfile
```diff
@@ -11,6 +11,6 @@ ENV VERSION ${version}
 COPY ./scripts/ /bin
 COPY ./src/main/resources/log4j2.properties /bin/log4j2.properties
 
-ADD target/hello-world-streams-1.0-SNAPSHOT.jar /
+ADD target/java-kafka-consumer-1.0-SNAPSHOT.jar /
 
-CMD ["/bin/run.sh", "/hello-world-streams-1.0-SNAPSHOT.jar"]
\ No newline at end of file
+CMD ["/bin/run.sh", "/java-kafka-consumer-1.0-SNAPSHOT.jar"]
```
scholzj: This looks wrong a well

scholzj (commented):
There seems to be lot of messy renamingd. Not sure if this is just the GitHub diff which messes things up or if it is wrong in the actual code. Please doublecheck.

java/kafka/consumer/Makefile
ppatierno: I took a look at all Makefile and Dockerfile and they are right. I really don't know why GitHub diff is showing this. Maybe renaming folders drove it crazy :-)

java/kafka/consumer/Dockerfile
ppatierno: Ditto as above

scholzj (approved):

PR #51, opened 2021-01-26
Add exception cause to producer warning, fix typo

im-konge:
Signed-off-by: Lukas Kral <lukywill16@gmail.com>

### DIFF
--- a/java/kafka/producer/src/main/java/KafkaProducerExample.java
+++ b/java/kafka/producer/src/main/java/KafkaProducerExample.java
@@ -63,7 +63,7 @@ public class KafkaProducerExample {
         AtomicLong numSent = new AtomicLong(0);
         for (long i = 0; i < config.getMessageCount(); i++) {
             if (transactionalProducer && i % msgPerTx == 0) {
-                log.info("Begging new transaction. Messages sent: {}", i);
+                log.info("Beginning new transaction. Messages sent: {}", i);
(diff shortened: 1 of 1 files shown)


java/kafka/producer/src/main/java/KafkaProducerExample.java
```diff
@@ -74,7 +74,7 @@ public static void main(String[] args) throws InterruptedException {
                     // Increment number of sent messages only if ack is received by producer
                     numSent.incrementAndGet();
                 } catch (ExecutionException e) {
-                    log.warn("Message {} wasn't sent properly!", i);
+                    log.warn("Message {} wasn't sent properly! \n{}", i, e.getCause());
```
tombentley: Probably no need for the `\n{}`. The methods of `log` print the exception and stacktrace for an "extra" exception argument (i.e. for which there isn't a `{}` placeholder). If it's a `KafkaException` there won't be a stacktrace, but if it's not there will be, so printing it with the stacktrace is worthwhile.

im-konge: Okay, thanks for the explanation :slightly_smiling_face:

tombentley (approved):

ppatierno (approved):

scholzj (approved):

PR #54, opened 2021-03-09
Parse additional config at the end of props setup to allow rewrite configurations

Frawless:
Signed-off-by: Jakub Stejskal <xstejs24@gmail.com>

This PR allows us to rewrite all configurations just only via `ADDITIONAL_CONFIG` value. This change will be very useful for integrating clients-examples into strimzi system tests. without this, we will need some additional env wars to setup scram-sha for example, etc.

Tests: tried few things manually and run some tests from the operator repo and everything looks good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,11 +63,12 @@ Producer
 * `USER_CRT` - the user's certificate
 * `USER_KEY` - the user's private key
 * `LOG_LEVEL` - logging level  
-* `PRODUCER_ACKS` = acknowledgement level
-* `HEADERS` = custom headers list separated by commas of `key1=value1, key2=value2`
-* `ADDITIONAL_CONFIG` = additional configuration for a producer application. The form is `key=value` records separated by new line character
-* `BLOCKING_PRODUCER` = if it's set, the producer will block another message until ack will be received
-* `MESSAGES_PER_TRANSACTION` = how many messages will be part of one transaction. Transaction config could be set via `ADDITIONAL_CONFIG` variable. Default is 10.
+* `PRODUCER_ACKS` - acknowledgement level
+* `HEADERS` - custom headers list separated by commas of `key1=value1, key2=value2`
+* `ADDITIONAL_CONFIG` - additional configuration for a producer application. The form is `key=value` records separated by new line character
+* `BLOCKING_PRODUCER` - if it's set, the producer will block another message until ack will be received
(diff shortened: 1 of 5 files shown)


ppatierno:
My doubt about this change is that it's then not clear what "additional configuration" means if you can not just adding but even overwrite something that is defined via other env var.

Frawless:
@ppatierno good point...on the other hand, the client will become too complex from my POV if we will add X another env vars and remove the option to overwrite them via `additionalConfig`. Maybe just rename it and leave note of why it's there?

michalxo:
I have took liberty of @Frawless idea (as I needed to pass some Sasl oauth plain stuff in it) and using additional Configuration seems like a nice way to sneak into config :-)

0) Main goal of my changes is to get rid of `sasl.login.callback.handler.class`, as it was used there hardwired. 
0) Using this additional configuration, sasl.mechanism can be overridden as well (hardwired to OAUTHBEARER now)
0) Minor updates:
  - some clean up code
  - simplify statements
  - added readme docs for this new env variable

michalxo:
I am having some problems with SASL after my changes, so please do not merge yet this.

UPD: Problems resolved in a just pushed commit. Ready for review (and hopefully merge).

README.md
```diff
@@ -63,11 +63,12 @@ Producer
...
 * `USER_KEY` - the user's private key
 * `LOG_LEVEL` - logging level  
-* `PRODUCER_ACKS` = acknowledgement level
-* `HEADERS` = custom headers list separated by commas of `key1=value1, key2=value2`
-* `ADDITIONAL_CONFIG` = additional configuration for a producer application. The form is `key=value` records separated by new line character
-* `BLOCKING_PRODUCER` = if it's set, the producer will block another message until ack will be received
-* `MESSAGES_PER_TRANSACTION` = how many messages will be part of one transaction. Transaction config could be set via `ADDITIONAL_CONFIG` variable. Default is 10.
+* `PRODUCER_ACKS` - acknowledgement level
+* `HEADERS` - custom headers list separated by commas of `key1=value1, key2=value2`
+* `ADDITIONAL_CONFIG` - additional configuration for a producer application. The form is `key=value` records separated by new line character
```
ppatierno, line 68: if we are using the ADDITIONAL_CONFIG for even overwriting the ones that can be set via other env vars (i.e. PRODUCER_ACKS) we should say that in the doc at least, otherwise changing the name of the env var to make it clearer but maybe breaking backward compatibility (even if it's just an example so we should be able to do that).

michalxo:
I am closing this PR in favor of #55

PR #58, opened 2021-05-02
Go Sarama examples

ppatierno:
This PR adds Golang examples using Sarama library.
I had this ready for so long time but struggling a little bit on adding Makefile for these examples. While I have something ready about that, I would like to add the source code examples false and procede with a PR for building afterwoads.


scholzj (approved):
It is compared to the other examples also missing a lot of configuration stuff (different authentication mechanisms etc.) and example YAMLs. I'm fine if you wanna add it later. But you should plan with that as well.

ppatierno:
Yeah that was the idea, it's a starting point. Even the other examples started to be simpler than what they are today.

PR #61, opened 2021-06-10
Update the OAuth library to 0.8.0

scholzj:
This PR updates the client examples to use the Strimzi OAuth library 0.8.0. Since 0.8.0 can load certificates directly from the PEM format, it uses this new feature and removes the old certificate preparation using JKS format. It also updates some other dependencies such as Kafka clients or tracing.

### DIFF
--- a/java/kafka/deployment-oauth.yaml
+++ b/java/kafka/deployment-oauth.yaml
@@ -88,7 +88,9 @@ spec:
                 key: clientSecret
           - name: OAUTH_TOKEN_ENDPOINT_URI
             value: https://sso.myproject.svc:8443/auth/realms/internal/protocol/openid-connect/token
-          - name: OAUTH_CRT
+          - name: OAUTH_SSL_TRUSTSTORE_TYPE
+            value: PEM
+          - name: OAUTH_SSL_TRUSTSTORE_CERTIFICATES
             valueFrom:
               secretKeyRef:
                 name: sso-x509-https-secret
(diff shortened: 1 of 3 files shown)


java/kafka/deployment-oauth.yaml
```diff
@@ -88,7 +88,9 @@ spec:
                 key: clientSecret
           - name: OAUTH_TOKEN_ENDPOINT_URI
             value: https://sso.myproject.svc:8443/auth/realms/internal/protocol/openid-connect/token
-          - name: OAUTH_CRT
+          - name: OAUTH_SSL_TRUSTSTORE_TYPE
+            value: PEM
```
ppatierno, line 92: quick question ... is there any other type other than PEM? I mean is it still possible to work as the previous way?

scholzj: The OAuth library supports PEM, PKCS12 and JKS formats. Only PEM is supported directly int he configuration, other types need to be provided a path to a file.

ppatierno: Ok so just to understand if I got it correctly, you deleted the part in the script that builds the truststore because now you can just provide, for example, a p12 file path in the `OAUTH_SSL_TRUSTSTORE_CERTIFICATES` if you specify PKCS12 as `OAUTH_SSL_TRUSTSTORE_TYPE`, is that right?

scholzj: No, the path to the PKCS12 would need to be specified in `OAUTH_SSL_TRUSTSTORE_LOCATION`. `OAUTH_SSL_TRUSTSTORE_CERTIFICATES` is only for having the content of the certificate set directly which is supported only with PEM files.

ppatierno (approved):

PR #84, opened 2022-10-18
Bump to Kafka version 3.3

OwenCorrigan76:
Bumped Kafka dependency to version 3.3.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,4 +109,4 @@ To enable tracing, configure the Jaeger Tracer using [environment variables](htt
 To run Jaeger Tracing, you can also use the provided example in [`deployment-tracing-jaeger.yaml`](./java/kafka/deployment-tracing-jaeger.yaml).
 To run Opentelemetry Tracing, you can also use the provided example in [`deployment-tracing-opentelemetry.yaml`](./java/kafka/deployment-tracing-opentelemetry.yaml).
 
-Jaeger / OpenTracing tracing is supported only in consumers / producers because OpenTracing support for Kafka Streams API is not compatible with latest Kafka version.
+Jaeger / OpenTracing tracing is supported only in consumers / producers because OpenTracing support for Kafka Streams API is not compatible with the latest Kafka versions.
--- a/README.md
+++ b/README.md
@@ -108,3 +108,5 @@ To enable tracing, configure the Jaeger Tracer using [environment variables](htt
 
 To run Jaeger Tracing, you can also use the provided example in [`deployment-tracing-jaeger.yaml`](./java/kafka/deployment-tracing-jaeger.yaml).
 To run Opentelemetry Tracing, you can also use the provided example in [`deployment-tracing-opentelemetry.yaml`](./java/kafka/deployment-tracing-opentelemetry.yaml).
+
+Jaeger / OpenTracing tracing is supported only in consumers / producers because OpenTracing support for Kafka Streams API is not compatible with latest Kafka version.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -108,3 +108,5 @@ To enable tracing, configure the Jaeger Tracer using [environment variables](htt
 
 To run Jaeger Tracing, you can also use the provided example in [`deployment-tracing-jaeger.yaml`](./java/kafka/deployment-tracing-jaeger.yaml).
 To run Opentelemetry Tracing, you can also use the provided example in [`deployment-tracing-opentelemetry.yaml`](./java/kafka/deployment-tracing-opentelemetry.yaml).
+
+*NOTE* : Jaeger Tracink eg is only supported up to Kafka version 2.8.0. For Kafka 2.8 +, use OpenTelemetry.
```
ppatierno: I thinks this should be removed, see @scholzj comment here as well https://github.com/strimzi/client-examples/pull/86#discussion_r1000747589

ppatierno (changes requested):

README.md
OwenCorrigan76: @ppatierno @scholzj @kyguy We know that Jaeger still works with consumer and producer but not with streams in version 3.3. Should I replace the comment in the README with something else or just remove completely?

scholzj: First of all, there is no versioning for the client examples. Just the source codes. So the reference to 2.8 makes no sense as there is no way to use the 2.8 version. Also, OpenTelemetry and OpenTracing do the same, but are not necessarily one to one replacement.

I don't think we should block the progress => so I would suggest to:
* Remove OpenTracing support from the Streams example
* Keep it in the Producer / Consumer examples where it IMHO still works fine
* Add a note to the README in this sense (=> Jaeger / OpenTracing tracing is supported only in consumers / producers because OPenTracing support for KAfka Streams API is not compatible with latest Kafka versions)

OwenCorrigan76: Thanks @scholzj. Changes made, tested and pushed.

ppatierno (changes requested):
I see some changes which are related to the NPE fix, they should not be there. You should remove them and rebase against the main where I merged your NPE fix already.

java/kafka/streams/src/main/java/KafkaStreamsExample.java
```diff
@@ -39,11 +39,7 @@ public static void main(String[] args) {
 
         TracingSystem tracingSystem = config.getTracingSystem();
         if (tracingSystem != TracingSystem.NONE) {
-            if (tracingSystem == TracingSystem.JAEGER) {
-
-                KafkaClientSupplier supplier = new  io.opentracing.contrib.kafka.streams.TracingKafkaClientSupplier(TracingInitializer.jaegerInitialize());
-                streams = new KafkaStreams(builder.build(), props, supplier);
-            } else if (tracingSystem == TracingSystem.OPENTELEMETRY) {
+            if (tracingSystem == TracingSystem.OPENTELEMETRY) {
```
ppatierno, line 42: I think you should also remove the Jaeger related dependencies in the Streams application pom.xml file.

ppatierno (changes requested):

kyguy (approved):

ppatierno (approved):
LGTM. Assuming the Streams application was tested with Jaeger and check that OpenTelemetry cannot be enabled. And that Producer and Consumer apps work properly with new Kafka 3.3.1

OwenCorrigan76:
@ppatierno Yes, I just tested this again. Kafka Version 3.3. Jaeger tracing working as expected with producer and consumer. Not working with streams. OpenTelemtry tracing producing traces with producer / consumer and streams.  
@scholzj

scholzj (approved):

scholzj:
Thanks for the PR.

Issue #85, opened 2022-10-19
Env variable TRACING_SYSTEM cannot be blank


HontoNoRoger:
We just got a crash looping container for the `quay.io/strimzi-examples/java-kafka-consumer:latest` image, as the newly introduced env variable `TRACING_SYSTEM` cannot be blank.

The error we see is:
```
Exception in thread "main" java.lang.NullPointerException
	at TracingSystem.forValue(TracingSystem.java:12)
	at KafkaConsumerConfig.fromEnv(KafkaConsumerConfig.java:75)
	at KafkaConsumerExample.main(KafkaConsumerExample.java:22)
```

It seems the tracing system needs to be set currently, otherwise the application won't run: https://github.com/strimzi/client-examples/blob/main/java/kafka/consumer/src/main/java/KafkaConsumerConfig.java#L75

scholzj:
@OwenCorrigan76 Can you please look at this?

ppatierno:
Closed via #86

PR #86, opened 2022-10-20
tracing nullPointerException fix

OwenCorrigan76:
Addressing the following issue: https://github.com/strimzi/client-examples/issues/85

### DIFF
--- a/java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
+++ b/java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
@@ -72,7 +72,7 @@ public class KafkaConsumerConfig {
         String oauthRefreshToken = System.getenv("OAUTH_REFRESH_TOKEN");
         String oauthTokenEndpointUri = System.getenv("OAUTH_TOKEN_ENDPOINT_URI");
         String additionalConfig = System.getenv().getOrDefault("ADDITIONAL_CONFIG", "");
-        TracingSystem tracingSystem = TracingSystem.forValue(System.getenv("TRACING_SYSTEM"));
+        TracingSystem tracingSystem = TracingSystem.forValue(System.getenv().getOrDefault("TRACING_SYSTEM", ""));
 
         return new KafkaConsumerConfig(bootstrapServers, topic, groupId, clientRack, messageCount, sslTruststoreCertificates, sslKeystoreKey,
                 sslKeystoreCertificateChain, oauthClientId, oauthClientSecret, oauthAccessToken, oauthRefreshToken, oauthTokenEndpointUri,
--- a/java/kafka/producer/src/main/java/KafkaProducerConfig.java
+++ b/java/kafka/producer/src/main/java/KafkaProducerConfig.java
@@ -77,7 +77,7 @@ public class KafkaProducerConfig {
         String acks = System.getenv().getOrDefault("PRODUCER_ACKS", "1");
         String headers = System.getenv("HEADERS");
         String additionalConfig = System.getenv().getOrDefault("ADDITIONAL_CONFIG", "");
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -108,3 +108,5 @@ To enable tracing, configure the Jaeger Tracer using [environment variables](htt
 
 To run Jaeger Tracing, you can also use the provided example in [`deployment-tracing-jaeger.yaml`](./java/kafka/deployment-tracing-jaeger.yaml).
 To run Opentelemetry Tracing, you can also use the provided example in [`deployment-tracing-opentelemetry.yaml`](./java/kafka/deployment-tracing-opentelemetry.yaml).
+
+*NOTE* : Jaeger Tracink eg is only supported up to Kafka version 2.8.0. For Kafka 2.8 +, use OpenTelemetry.
```
scholzj: I do not think this is true and should be here.

java/kafka/consumer/src/main/java/KafkaConsumerExample.java
```diff
@@ -27,43 +27,45 @@ public static void main(String[] args) {
         int receivedMsgs = 0;
 
         TracingSystem tracingSystem = config.getTracingSystem();
-        if (tracingSystem != TracingSystem.NONE) {
-            if (tracingSystem == TracingSystem.JAEGER) {
-                TracingInitializer.jaegerInitialize();
+        if (tracingSystem != null) {
```
scholzj: When would this be null? Also, why if this is null do we skip the messages? I would expect this to be either NONE or JAEGER or OPENTELEMETRY.

java/kafka/consumer/src/main/java/KafkaConsumerExample.java
```diff
@@ -27,43 +27,45 @@ public static void main(String[] args) {
...
+                } else if (tracingSystem == TracingSystem.OPENTELEMETRY) {
+                    TracingInitializer.otelInitialize();
 
-                props.put(ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG, io.opentelemetry.instrumentation.kafkaclients.TracingConsumerInterceptor.class.getName());
-            } else {
-                log.error("Error: TRACING_SYSTEM {} is not recognized or supported!", tracingSystem);
+                    props.put(ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG, io.opentelemetry.instrumentation.kafkaclients.TracingConsumerInterceptor.class.getName());
+                } else {
+                    log.error("Error: TRACING_SYSTEM {} is not recognized or supported!", tracingSystem);
+                }
```
scholzj: Thsi should be just one if-else -> you should not nest it.

java/kafka/pom.xml
```diff
@@ -18,7 +18,7 @@
         <maven.compiler.target>11</maven.compiler.target>
         <log4j.version>2.17.1</log4j.version>
         <slf4j-simple.version>1.6.2</slf4j-simple.version>
-        <kafka.version>2.8.0</kafka.version>
+        <kafka.version>3.3.1</kafka.version>
```
scholzj: You seem to be mixing two different PRs.

java/kafka/producer/src/main/java/KafkaProducerExample.java
```diff
@@ -31,6 +31,7 @@ public static void main(String[] args) throws InterruptedException {
         List<Header> headers = null;
 
         TracingSystem tracingSystem = config.getTracingSystem();
+       // if (TracingSystem == null)
```
scholzj: Can be deleted?

scholzj (approved):
Thanks

ppatierno (approved):
LGTM. Thanks!

PR #87, opened 2022-11-01
further remove jaeger from StreamsExaxmple

OwenCorrigan76:
Changed 2 things:

- Removed Jaeger / openTracing from Streams in deployment-tracing-jaeger.yaml
- Throw RuntimeError when jaeger is configured for Streams application

### DIFF
--- a/java/kafka/streams/src/main/java/KafkaStreamsExample.java
+++ b/java/kafka/streams/src/main/java/KafkaStreamsExample.java
@@ -46,7 +46,7 @@ public class KafkaStreamsExample {
                 KafkaClientSupplier supplier = new TracingKafkaClientSupplier();
                 streams = new KafkaStreams(builder.build(), props, supplier);
             } else {
-                throw new RuntimeException("Error: TRACING_SYSTEM jaeger is not recognized or supported!");
+                throw new RuntimeException("Error: TRACING_SYSTEM " + tracingSystem + " is not recognized or supported!");
             }
         } else {
             streams = new KafkaStreams(builder.build(), props);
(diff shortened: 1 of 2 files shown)


scholzj (approved):

kyguy (approved):

java/kafka/streams/src/main/java/KafkaStreamsExample.java
```diff
@@ -46,8 +46,7 @@ public static void main(String[] args) {
                 KafkaClientSupplier supplier = new TracingKafkaClientSupplier();
                 streams = new KafkaStreams(builder.build(), props, supplier);
             } else {
-                log.error("Error: TRACING_SYSTEM {} is not recognized or supported!", config.getTracingSystem());
-                streams = new KafkaStreams(builder.build(), props);
+                throw new RuntimeException("Error: TRACING_SYSTEM jaeger is not recognized or supported!");
```
ppatierno: because it's in the `else` you are not saying just that "jaeger" is not supported, you are saying that everything else different from "opentelemetry" is not supported, so you should change the error message accordingly. Something that shows what's the content of `tracingSystem` variable and saying it's not recognized or supported.

ppatierno (changes requested):

java/kafka/streams/src/main/java/KafkaStreamsExample.java
OwenCorrigan76: @ppatierno  Refactord

ppatierno (approved):
LGTM. Thanks!

scholzj:
Thanks for the PR.

PR #91, opened 2022-12-06
Refactor Client-Example ConsumerConfigs

OwenCorrigan76:
Refactor the Producer, Consumer and Streams Config files in the Java client by reducing the complexity and size of the class. Standardise the naming scheme of environment variables used to configure the clients / standardise the envVars used to configure the Kafka client properties as outlined here [1].

Example of standardised Environmental Variables:
``` 
KAFKA_<NAME_OF_KAFKA_PROPERTY_DELIMITED_BY_UNDERSCORES>
```

[1]  https://github.com/kyguy/proposals/blob/refactor-client-examples/040-refactor-client-examples.md


scholzj (commented):
I know this is just a draft, so not sure what else do you plan. But I think you should go deeper ... I do not think there is any need to have some special treatment for the TLS certificates or OAuth authentication. I think they can be configured from the `KAFKA_...` env vars as well.

java/kafka/producer/src/main/java/KafkaProducerConfig.java
```diff
@@ -210,23 +210,23 @@ public TracingSystem getTracingSystem() {
...
+                ", oauthClientId='" + oauthClientId + '\'' +
+                ", oauthClientSecret='" + oauthClientSecret + '\'' +
+                ", oauthAccessToken='" + oauthAccessToken + '\'' +
+                ", oauthRefreshToken='" + oauthRefreshToken + '\'' +
+                ", oauthTokenEndpointUri='" + oauthTokenEndpointUri + '\'' +
+                ", additionalConfig='" + additionalConfig + '\'' +
+                ", tracingSystem='" + tracingSystem + '\'' +
+                '}';
     }
 }
```
kyguy: Undo changes

java/kafka/streams/src/main/java/KafkaStreamsExample.java
```diff
@@ -46,8 +46,7 @@ public static void main(String[] args) {
                 KafkaClientSupplier supplier = new TracingKafkaClientSupplier();
                 streams = new KafkaStreams(builder.build(), props, supplier);
             } else {
-                log.error("Error: TRACING_SYSTEM {} is not recognized or supported!", config.getTracingSystem());
-                streams = new KafkaStreams(builder.build(), props);
+                throw new RuntimeException("Error: TRACING_SYSTEM " + tracingSystem + " is not recognized or supported!");
             }
         } else {
             streams = new KafkaStreams(builder.build(), props);
```
kyguy: undo changes

java/kafka/deployment-tracing-jaeger.yaml
```diff
@@ -61,47 +61,6 @@ spec:
...
-              value: my-jaeger-agent
-            - name: JAEGER_SAMPLER_TYPE
-              value: const
-            - name: JAEGER_SAMPLER_PARAM
-              value: "1"
-            - name: TRACING_SYSTEM
-              value: jaeger
----
-apiVersion: apps/v1
-kind: Deployment
```
kyguy: Undo changes

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
```diff
@@ -10,175 +10,121 @@
...
-                String key = record.substring(0, endIndex);
-                String value = record.substring(endIndex + 1);
-                additionalProps.put(key.trim(), value.trim());
-            }
+            props.putAll(DEFAULT_KEYSTORE_CONFIGS);
         }
 
         if ((config.getOauthAccessToken() != null)
                 || (config.getOauthTokenEndpointUri() != null && config.getOauthClientId() != null && config.getOauthRefreshToken() != null)
-                || (config.getOauthTokenEndpointUri() != null && config.getOauthClientId() != null && config.getOauthClientSecret() != null))    {
```
kyguy, line 126: undo indentation change

README.md
```diff
@@ -76,16 +76,13 @@ Producer
...
-* `CA_CRT` - the certificate of the CA which signed the brokers' TLS certificates, for adding to the client's trust store
-* `USER_CRT` - the user's certificate
-* `USER_KEY` - the user's private key
-* `LOG_LEVEL` - logging level  
+* `KAFKA_BOOTSTRAP_SERVERS` - comma-separated host and port pairs that is a list of Kafka broker addresses. The form of pair is `host:port`, e.g. `my-cluster-kafka-bootstrap:9092` 
+* `STRIMZI_TOPIC` - name of topic which consumer subscribes  
+* `KAFKA_GROUP_ID` - specifies the consumer group id for the consumer
+* `STRIMZI_MESSAGE_COUNT` - the number of messages the consumer should receive
+* `STRIMZI_LOG_LEVEL` - logging level  
 * `ADDITIONAL_CONFIG` - additional configuration for a consumer application. Notice, that you can also override any previously set variable by setting this. The form is `key=value` records separated by new line character
```
kyguy: This can be removed

README.md
```diff
@@ -76,16 +76,13 @@ Producer
...
-* `TOPIC` - name of topic which consumer subscribes  
-* `GROUP_ID` - specifies the consumer group id for the consumer
-* `MESSAGE_COUNT` - the number of messages the consumer should receive
-* `CA_CRT` - the certificate of the CA which signed the brokers' TLS certificates, for adding to the client's trust store
-* `USER_CRT` - the user's certificate
-* `USER_KEY` - the user's private key
-* `LOG_LEVEL` - logging level  
+* `KAFKA_BOOTSTRAP_SERVERS` - comma-separated host and port pairs that is a list of Kafka broker addresses. The form of pair is `host:port`, e.g. `my-cluster-kafka-bootstrap:9092` 
+* `STRIMZI_TOPIC` - name of topic which consumer subscribes  
+* `KAFKA_GROUP_ID` - specifies the consumer group id for the consumer
```
kyguy: Env vars used for Kafka client configuration can be removed.

I would also add a note to the README.md explaining that all Kafka client properties can be configured by adding  an env var with the Kafka property name in a capital letters, delimited by "_", and prefixed with "KAFKA"

OwenCorrigan76:
@scholzj When you say "special treatment", do you mean these lines [1] [2]? I kept these in to give the same behaviour as the code before it. Are you saying we can strip this code out completely? If so, are we removing some of the assistance we're offering to the user? So, the user will have to make sure Ssl. Oauth etc. are configured properly from, their end? Thanks

[1] https://github.com/strimzi/client-examples/blob/e40023abb5b7c5d3e1be8a038dca060092e22bde/java/kafka/consumer/src/main/java/KafkaConsumerConfig.java#L90-L99

[2] https://github.com/strimzi/client-examples/blob/e40023abb5b7c5d3e1be8a038dca060092e22bde/java/kafka/consumer/src/main/java/KafkaConsumerConfig.java#L101-L111

scholzj:
Yes, that is the code I mean. Also the (de)serializers for example. You can call it an _assistance_. But it also makes it unclear how exactly it is configured. Because one day when they need to write their own app, they will need to dig into the code to understand how to actually do it. I think anything what ends up in the consumer properties file should go directly through the `KAFKA_` environment variables. The only part which should remain are the once which don't end up there ... e.g. number of messages, tracer initiliazation etc.

OwenCorrigan76:
@scholzj  @kyguy  I've removed the Ssl and oAuth configs and only left the USER_CONFIGS in. These are the ones that will have a STRIMZI_ prefix. If you guys give me the go ahead that this is the way to go, I'll address the other comments @kyguy left and then move on to the Producer and Streams. Thanks guys.

README.md
scholzj: You can remove this as well I guess.

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
```diff
@@ -3,228 +3,85 @@
...
-            ", oauthRefreshToken='" + oauthRefreshToken + '\'' +
-            ", oauthTokenEndpointUri='" + oauthTokenEndpointUri + '\'' +
-            ", additionalConfig='" + additionalConfig + '\'' +
-            ", tracingSystem='" + tracingSystem + '\'' +
-            '}';
+                "topic='" + topic + '\'' +
+                ", enableAutoCommit='" + enableAutoCommit + '\'' +
+                ", messageCount=" + messageCount +
+                ", tracingSystem='" + tracingSystem + '\'' +
+                kafkaConfigOptionsToString() +
```
scholzj: Is this needed? The Kafka client prints its config at startup. If you want to keep this, you should check the formatting to fit into the other fields.

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
```diff
@@ -3,228 +3,85 @@
...
-    public KafkaConsumerConfig(String bootstrapServers, String topic, String groupId, String clientRack, Long messageCount,
-                               String sslTruststoreCertificates, String sslKeystoreKey, String sslKeystoreCertificateChain,
-                               String oauthClientId, String oauthClientSecret, String oauthAccessToken, String oauthRefreshToken,
-                               String oauthTokenEndpointUri, String additionalConfig, TracingSystem tracingSystem) {
-        this.bootstrapServers = bootstrapServers;
+    private static final Map<String, String> USER_CONFIGS = System.getenv()
+            .entrySet()
+            .stream()
+            .filter(map -> map.getKey().startsWith(KAFKA_PREFIX))
+            .collect(Collectors.toMap(map -> convertEnvVarToPropertyKey(map.getKey()), map -> map.getValue()));
```
scholzj: 1) Should you use Properties directly here?
2) I also wonder if this should be static. In general, for things such as testing etc., it would be better to handle this in the constructor or in the `fromEnv` method. You might nto test it today, but maybe you can add some tests for this in the future.

java/kafka/producer/src/main/java/KafkaProducerConfig.java
```diff
@@ -210,23 +210,23 @@ public TracingSystem getTracingSystem() {
...
+                ", sslTruststoreCertificates='" + sslTruststoreCertificates + '\'' +
+                ", sslKeystoreKey='" + sslKeystoreKey + '\'' +
+                ", sslKeystoreCertificateChain='" + sslKeystoreCertificateChain + '\'' +
+                ", oauthClientId='" + oauthClientId + '\'' +
+                ", oauthClientSecret='" + oauthClientSecret + '\'' +
+                ", oauthAccessToken='" + oauthAccessToken + '\'' +
+                ", oauthRefreshToken='" + oauthRefreshToken + '\'' +
+                ", oauthTokenEndpointUri='" + oauthTokenEndpointUri + '\'' +
+                ", additionalConfig='" + additionalConfig + '\'' +
+                ", tracingSystem='" + tracingSystem + '\'' +
```
scholzj: Maybe this should not be part of this PR? Ideally, you should try to keep the PRs focused on a single thing.

java/kafka/streams/src/main/java/KafkaStreamsExample.java
```diff
@@ -46,8 +46,7 @@ public static void main(String[] args) {
                 KafkaClientSupplier supplier = new TracingKafkaClientSupplier();
                 streams = new KafkaStreams(builder.build(), props, supplier);
             } else {
-                log.error("Error: TRACING_SYSTEM {} is not recognized or supported!", config.getTracingSystem());
-                streams = new KafkaStreams(builder.build(), props);
+                throw new RuntimeException("Error: TRACING_SYSTEM " + tracingSystem + " is not recognized or supported!");
```
scholzj: Same as above ... Maybe this should not be part of this PR? Ideally, you should try to keep the PRs focused on a single thing.

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
```diff
@@ -3,228 +3,85 @@
...
-    private static final Logger log = LogManager.getLogger(KafkaConsumerConfig.class);
-
     private static final long DEFAULT_MESSAGES_COUNT = 10;
-    private final String bootstrapServers;
+    private static final String KAFKA_PREFIX = "KAFKA_";
+
     private final String topic;
-    private final String groupId;
-    private final String autoOffsetReset = "earliest";
     private final String enableAutoCommit = "false";
```
scholzj: This seems to be a weird left-over. You do not set it, but you keep it in the code. If it is needed, maybe you should check if it is enabled or not through the Kafka properties and set it from there? (https://kafka.apache.org/documentation/#consumerconfigs_enable.auto.commit)

README.md
```diff
@@ -76,16 +76,13 @@ Producer
...
+* `STRIMZI_MESSAGE_COUNT` - the number of messages the consumer should receive
+* `STRIMZI_LOG_LEVEL` - logging level
```
scholzj: You should expans this a bit. List the Strimzi specific options and elaborate that any Kafka otpions can be used as Env var by:
* Preficing it with `KAFKA_` and making it uppercase and using `_` instead of `.`.

scholzj (commented):
Some more comments, but it starts looking better. You should also update all the YAML example files to make it work with the new code.

README.md
OwenCorrigan76: Added:
Below are listed and described environmental variables that are NOT Kafka-specific but must be used. These environmental variables should be prefixed
with uppercase STRIMZI_  using _ instead of . (eg. STRIMZI_TOPIC). Any Kafka configuration option can be used as an environmental variable and should be prefixed with uppercase KAFKA_ using _ instead of . (eg. KAFKA_BOOTSTRAP_SERVERS).

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
OwenCorrigan76: autoOffsetReset removed. Discuss enableAutoCommit with @kyguy

java/kafka/streams/src/main/java/KafkaStreamsExample.java
OwenCorrigan76: Will remove

java/kafka/producer/src/main/java/KafkaProducerConfig.java
OwenCorrigan76: Will remove

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
OwenCorrigan76: Discuss today with @kyguy

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
OwenCorrigan76: Not needed now as we have ` kafkaConfigOptionsToString()`. Removed

README.md
OwenCorrigan76: @scholzj I've removed ADDITIONAL_CONFIG as it's no longer part of ConsumerConfig. Do you suggest removing the STRIMZI_ prefixed envVar examples too and just leave in the explanation at the beginning of the ##Configuration paragraph?

OwenCorrigan76: .yaml files updated with correct prefixes.

README.md
OwenCorrigan76: Done

README.md
OwenCorrigan76: Including the STRIMZI_ prefixed ones?

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
OwenCorrigan76: Full block of code removed. All security is being handled by the user via the envVars.

java/kafka/deployment-tracing-jaeger.yaml
OwenCorrigan76: Undone (should be correct now)

java/kafka/producer/src/main/java/KafkaProducerConfig.java
OwenCorrigan76: Fixed

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
scholzj: I think it is a bit more complicated then just removing it. The auto-commit is part of Kafka. But what behaviour do you expect when it is disabled? How does the client do the commits etc.? That is something you need to consider.

java/kafka/consumer/src/main/java/KafkaConsumerExample.java
```diff
@@ -41,7 +43,7 @@ public static void main(String[] args) {
             }
         }
 
-        boolean commit = !Boolean.parseBoolean(config.getEnableAutoCommit());
+       boolean commit = !Boolean.parseBoolean (props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
```
kyguy:
```suggestion
       boolean commit = !Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
```

java/kafka/deployment-ssl.yaml
```diff
@@ -110,18 +110,18 @@ spec:
...
+          - name: KAFKA_BOOTSTRAP_SERVERS
             value: my-cluster-kafka-bootstrap:9093
-          - name: TOPIC
+          - name: STRIMZI_TOPIC
             value: my-topic-reversed
-          - name: GROUP_ID
+          - name: KAFKA_GROUP_ID
             value: java-kafka-consumer
-          - name: LOG_LEVEL
+          - name: STRIMZI_ LOG_LEVEL
```
kyguy:
```suggestion
          - name: STRIMZI_LOG_LEVEL
```

java/kafka/deployment-tracing-jaeger.yaml
```diff
@@ -3,7 +3,7 @@ kind: KafkaTopic
 metadata:
   name: my-topic
   labels:
-      strimzi.io/cluster: my-cluster
+    strimzi.io/cluster: my-cluster
```
kyguy: extra two spaces removed

java/kafka/deployment-tracing-jaeger.yaml
```diff
@@ -117,26 +76,26 @@ spec:
...
+          - name: JAEGER_SERVICE_NAME
+            value: java-kafka-consumer
+          - name: JAEGER_AGENT_HOST
+            value: my-jaeger-agent
+          - name: JAEGER_SAMPLER_TYPE
+            value: const
+          - name: JAEGER_SAMPLER_PARAM
+            value: "1"
+          - name: TRACING_SYSTEM
+            value: jaeger
```
kyguy: add two space indentation back

java/kafka/streams/src/main/java/KafkaStreamsExample.java
```diff
@@ -47,8 +47,7 @@ public static void main(String[] args) {
                 streams = new KafkaStreams(builder.build(), props, supplier);
             } else {
                 log.error("Error: TRACING_SYSTEM {} is not recognized or supported!", config.getTracingSystem());
-                streams = new KafkaStreams(builder.build(), props);
-            }
+                streams = new KafkaStreams(builder.build(), props);   }
```
kyguy, line 50: Rogue formatting change, can be undone

java/kafka/streams/src/main/java/KafkaStreamsExample.java
```diff
@@ -47,8 +47,7 @@ public static void main(String[] args) {
                 streams = new KafkaStreams(builder.build(), props, supplier);
             } else {
                 log.error("Error: TRACING_SYSTEM {} is not recognized or supported!", config.getTracingSystem());
-                streams = new KafkaStreams(builder.build(), props);
-            }
```
kyguy: Rogue formatting change, can be undone

kyguy (commented):
Just some formatting nits

README.md
```diff
@@ -57,7 +57,9 @@ To run the OAuth example, you will need to have your Kafka cluster configured wi
 ## Configuration
 
 Although this Hello World is simple example it is fully configurable.
-Below are listed and described environmental variables.
+Below are listed and described environmental variables that are NOT Kafka-specific but must be used. 
+These environmental variables should be prefixed with `STRIMZI_` using `_` instead of `.` (eg. STRIMZI_TOPIC).
```
scholzj: These are just random environment variables. The `_` versus `.` makes no sense here. It makes sense only for the `KAFKA_` env vars which map to actual Kafka configuration properties.

README.md
```diff
@@ -57,7 +57,9 @@ To run the OAuth example, you will need to have your Kafka cluster configured wi
 ## Configuration
 
 Although this Hello World is simple example it is fully configurable.
-Below are listed and described environmental variables.
+Below are listed and described environmental variables that are NOT Kafka-specific but must be used. 
+These environmental variables should be prefixed with `STRIMZI_` using `_` instead of `.` (eg. STRIMZI_TOPIC). 
+Any Kafka configuration option can be given as an environmental variable and should be prefixed with `KAFKA_` using `_` instead of `.` (eg. bootstrap.servers -> KAFKA_BOOTSTRAP_SERVERS).
```
scholzj:
```suggestion
Any Kafka configuration option can be given as an environmental variable and should be prefixed with `KAFKA_` using `_` instead of `.` (eg. `bootstrap.servers` -> `KAFKA_BOOTSTRAP_SERVERS`).
```

scholzj: So, does this currently apply to consumers only? It is fine to do it step by step. But you should make it clear here in the README. Right now it sounds like this applies for producers and Streams as well.

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
```diff
@@ -3,228 +3,56 @@
...
-        // override properties with defined additional properties
-        props.putAll(additionalProps);
-
-        return props;
-    }
-
-    public String getBootstrapServers() {
-        return bootstrapServers;
+    public static String convertEnvVarToPropertyKey(String envVar) {
+        System.out.println("ENV_VAR is " + envVar);
```
scholzj: This should use logger if you want to log this (doesn't seem needed to me, but why not) or be removed if it is some leftover from debugging.

scholzj (commented):
Left few nits mostl around formatting etc. But looks good overall.

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
OwenCorrigan76: Removed altogether

README.md
OwenCorrigan76: Refactored to specify Consumer only (until later updates)

OwenCorrigan76: Added back in

README.md
OwenCorrigan76: Removed and refactored to fit the uniformity of the rest of the README.

OwenCorrigan76:
@scholzj @kyguy Thanks for the reviews / feedback. I've made the latest changes as commented by @scholzj. I will now start to refactor the ProducerConfigs. As I am doing so, I think it prudent to refactor the name of the PR to Refactor Client-Example Configs.

README.md
```diff
@@ -57,7 +57,10 @@ To run the OAuth example, you will need to have your Kafka cluster configured wi
 ## Configuration
 
 Although this Hello World is simple example it is fully configurable.
-Below are listed and described environmental variables.
+Below are listed and described environmental variables.
```
ppatierno: I think we use "environment variable" everywhere in our docs. Can we change this way everywhere you are using "environmental"?

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
```diff
@@ -3,228 +3,55 @@
...
-                additionalConfig, tracingSystem);
+        String topic = System.getenv("STRIMZI_TOPIC");
+        Long messageCount = System.getenv("STRIMZI_MESSAGE_COUNT") == null ? DEFAULT_MESSAGES_COUNT : Long.parseLong(System.getenv("STRIMZI_MESSAGE_COUNT"));
+        TracingSystem tracingSystem = TracingSystem.forValue(System.getenv().getOrDefault("STRIMZI_TRACING_SYSTEM", ""));
+        Properties properties = new Properties();
+        properties.putAll(System.getenv()
+                .entrySet()
+                .stream()
+                .filter(map -> map.getKey().startsWith(KAFKA_PREFIX))
+                .collect(Collectors.toMap(map -> convertEnvVarToPropertyKey(map.getKey()), map -> map.getValue())));
```
ppatierno: is `map` appropriate as variable here? You are getting a `Map.Entry` instance right? I would be more for naming it `entry` or `mapEntry` but not just `map` which makes reader thinking about the whole map itself.

java/kafka/consumer/src/main/java/KafkaConsumerExample.java
```diff
@@ -41,7 +43,7 @@ public static void main(String[] args) {
             }
         }
 
-        boolean commit = !Boolean.parseBoolean(config.getEnableAutoCommit());
+       boolean commit = !Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
```
ppatierno: wrong indentation, missing space?

scholzj:
> @scholzj @kyguy Thanks for the reviews / feedback. I've made the latest changes as commented by @scholzj. I will now start to refactor the ProducerConfigs. As I am doing so, I think it prudent to refactor the name of the PR to Refactor Client-Example Configs.

Too late I guess. But we should have first approved and merged the consumer config you had. This just makes it harder to review and fix things.

java/kafka/producer/src/main/java/KafkaProducerConfig.java
```diff
@@ -3,230 +3,85 @@
...
-            }
-        }
-
-        // override properties with defined additional properties
-        props.putAll(additionalProps);
-
-        return props;
+        String topic = System.getenv("STRIMZI_TOPIC");
+        Long messageCount = System.getenv("STRIMZI_MESSAGE_COUNT") == null ? DEFAULT_MESSAGES_COUNT : Long.parseLong(System.getenv("STRIMZI_MESSAGE_COUNT"));
+        int delay = System.getenv("STRIMZI_DELAY") == null ? DEFAULT_DELAY : Integer.parseInt(System.getenv("STRIMZI_DELAY"));
```
kyguy:
```suggestion
        int delay = Integer.parseInt(System.getenv("STRIMZI_DELAY_MS"));
```

java/kafka/producer/src/main/java/KafkaProducerConfig.java
```diff
@@ -3,230 +3,85 @@
...
-import java.util.StringTokenizer;
+import java.util.stream.Collectors;
 
 public class KafkaProducerConfig {
     private static final Logger log = LogManager.getLogger(KafkaProducerConfig.class);
 
     private static final long DEFAULT_MESSAGES_COUNT = 10;
     private static final String DEFAULT_MESSAGE = "Hello world";
-    private final String bootstrapServers;
+    private static final int DEFAULT_DELAY = 1;
```
kyguy: Can be removed

java/kafka/producer/src/main/java/KafkaProducerExample.java
```diff
@@ -19,6 +19,7 @@
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicLong;
 
+
```
kyguy: remove

java/kafka/producer/src/main/java/KafkaProducerExample.java
```diff
@@ -57,8 +58,11 @@ public static void main(String[] args) throws InterruptedException {
         KafkaProducer producer = new KafkaProducer(props);
         log.info("Sending {} messages ...", config.getMessageCount());
 
+        // Address
         boolean blockProducer = System.getenv("BLOCKING_PRODUCER") != null;
-        boolean transactionalProducer = config.getAdditionalConfig().contains("transactional.id");
+      //  boolean transactionalProducer = !Boolean.parseBoolean(props.getProperty(TRANSACTIONAL_ID_CONFIG));
+        boolean transactionalProducer = config.getProperties().contains("transactional.id");
+      //  boolean transactionalProducer = config.getAdditionalConfig().contains("transactional.id");
```
kyguy: Don't forget to clean up

java/kafka/streams/src/main/java/KafkaStreamsExample.java
kyguy: Revert to original

java/kafka/streams/src/main/java/KafkaStreamsExample.java
```diff
@@ -46,13 +46,12 @@ public static void main(String[] args) {
                 KafkaClientSupplier supplier = new TracingKafkaClientSupplier();
                 streams = new KafkaStreams(builder.build(), props, supplier);
             } else {
-                log.error("Error: TRACING_SYSTEM {} is not recognized or supported!", config.getTracingSystem());
-                streams = new KafkaStreams(builder.build(), props);
```
kyguy: Add back in, revert to original

java/kafka/streams/src/main/java/KafkaStreamsExample.java
```diff
@@ -54,4 +54,4 @@ public static void main(String[] args) {
 
         streams.start();
     }
-}
```
kyguy: Undo

kyguy (approved):

ppatierno:
@OwenCorrigan76 can we move out of the draft state?

ppatierno:
Agree with @scholzj ... but it seems not to be late yet. There are no changes on the producer side yet :-)

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
```diff
@@ -3,228 +3,56 @@
...
-
-        // override properties with defined additional properties
-        props.putAll(additionalProps);
-
-        return props;
-    }
-
-    public String getBootstrapServers() {
-        return bootstrapServers;
+    public static String convertEnvVarToPropertyKey(String envVar) {
```
scholzj: Can be private I guess?

README.md
```diff
@@ -57,7 +57,10 @@ To run the OAuth example, you will need to have your Kafka cluster configured wi
 ## Configuration
 
 Although this Hello World is simple example it is fully configurable.
-Below are listed and described environmental variables.
+Below are listed and described environment variables. 
+Regarding Consumer Only (until later updates): below are listed and described environmental variables that are NOT Kafka-specific but must be used.
```
scholzj: Only with lowercase `o` I guess? Also, is `must be used` the right way?

README.md
```diff
@@ -57,7 +57,10 @@ To run the OAuth example, you will need to have your Kafka cluster configured wi
 ## Configuration
```
scholzj, line 57: Maybe you can do something like this (just the `Configuration` section):

```markdown
## Configuration

Below are listed and described the available environment variables that can be used for configuration. 

### Producer  

* `BOOTSTRAP_SERVERS` - comma-separated host and port pairs that is a list of Kafka broker addresses. The form of pair is `host:port`, e.g. `my-cluster-kafka-bootstrap:9092` 
* `TOPIC` - the topic the producer will send to  
* `DELAY_MS` - the delay, in ms, between messages  
* `MESSAGE_COUNT` - the number of messages the producer should send  
* `CA_CRT` - the certificate of the CA which signed the brokers' TLS certificates, for adding to the client's trust store
* `USER_CRT` - the user's certificate
* `LOG_LEVEL` - logging level  
* `PRODUCER_ACKS` - acknowledgement level
* `HEADERS` - custom headers list separated by commas of `key1=value1, key2=value2`
* `BLOCKING_PRODUCER` - if it's set, the producer will block another message until ack will be received
* `MESSAGES_PER_TRANSACTION` - how many messages will be part of one transaction. Transaction config could be set via `ADDITIONAL_CONFIG` variable. Default is 10.
* `ADDITIONAL_CONFIG` - additional configuration for a producer application. Notice, that you can also override any previously set variable by setting this. The form is `key=value` records separated by new line character
* `TRACING_SYSTEM` - if it's set to `jaeger` or `opentelemetry`, this will enable tracing. 

### Consumer  

* `STRIMZI_TOPIC` - name of topic to which consumer subscribes  
* `STRIMZI_MESSAGE_COUNT` - the number of messages the consumer should receive
* `STRIMZI_LOG_LEVEL` - logging level  
* `STRIMZI_TRACING_SYSTEM` - if it's set to `jaeger` or `opentelemetry`, this will enable tracing.

Additionally, any Kafka Consumer API configuration option can be passed as an environmental variable.
It should be prefixed with `KAFKA_` and use `_` instead of `.`.
For example environment variable `KAFKA_BOOTSTRAP_SERVERS` will be used as the `bootstrap.servers` configuration option in the Kafka Consumer API.

### Streams  

* `BOOTSTRAP_SERVERS` - comma-separated host and port pairs that is a list of Kafka broker addresses. The form of pair is `host:port`, e.g. `my-cluster-kafka-bootstrap:9092`
* `APPLICATION_ID` - The Kafka Streams application ID
* `SOURCE_TOPIC` - name of topic which will be used as the source of messages
* `TARGET_TOPIC` - name of topic where the transformed images are sent
* `COMMIT_INTERVAL_MS` - the interval for the Kafka Streams consumer part committing the offsets
* `CA_CRT` - the certificate of the CA which signed the brokers' TLS certificates, for adding to the client's trust store
* `USER_CRT` - the user's certificate
* `LOG_LEVEL` - logging level
* `ADDITIONAL_CONFIG` - additional configuration for a streams application. Notice, that you can also override any previously set variable by setting this. The form is `key=value` records separated by new line character.
* `TRACING_SYSTEM` - if it's set to `jaeger` or `opentelemetry`, this will enable tracing.

### Tracing

The examples support tracing using the [OpenTracing Apache Kafka Instrumentation](https://github.com/opentracing-contrib/java-kafka-client), 
[OpenTelemetry Java Instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation) and the [Jaeger project](https://www.jaegertracing.io/).
To enable tracing, configure the Jaeger Tracer using [environment variables](https://github.com/jaegertracing/jaeger-client-java/tree/master/jaeger-core#configuration-via-environment).

To run OpenTelemetry Tracing, you can also use the provided example in [`deployment-tracing-opentelemetry.yaml`](./java/kafka/deployment-tracing-opentelemetry.yaml).

Jaeger / OpenTracing tracing is supported only in consumers / producers because OpenTracing support for Kafka Streams API is not compatible with the latest Kafka versions.
```

java/kafka/streams/src/main/java/KafkaStreamsExample.java
OwenCorrigan76: Done

java/kafka/producer/src/main/java/KafkaProducerExample.java
OwenCorrigan76: Reverted to before refactoring ConsumerConfig

java/kafka/consumer/src/main/java/KafkaConsumerExample.java
OwenCorrigan76: Fixed

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
OwenCorrigan76: Refactored to use mapEntry

README.md
OwenCorrigan76: Done

OwenCorrigan76:
@scholzj @ppatierno I had done some work on ProducerConfigs but I reverted everything back to before I changed ProducerConfigs. I guess I'll concentrate on Consumer first. Get it merged and then open a new PR for Producer. When that is merged, I'll open a separate PR for Streams. This means that any of Kyle's comments regarding ProducerConfigs above can be ignored. I know it's messy but i originally thought you guys would prefer Producer / Consumer and Streams done in the same PR (my mistake).

ppatierno (approved):

java/kafka/consumer/src/main/java/KafkaConsumerConfig.java
OwenCorrigan76: Refactored

README.md
OwenCorrigan76: Refactored

README.md
OwenCorrigan76: Makes sense. Refactored to use this suggestion.

scholzj (approved):
Thanks for the PR!

PR #94, opened 2022-12-23
Revert "Refactor Client-Example ConsumerConfigs"

scholzj:
There seem to be many issues with how the code introduced in #91 actually works - not as much in the code itself but in the example YAMLs. They are missing deserializers, the TLS configuration is completely wrong etc. This PR reverts that PR - we should get back to it, fix the YAMLs and test them properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,8 @@ To run the OAuth example, you will need to have your Kafka cluster configured wi
 
 ## Configuration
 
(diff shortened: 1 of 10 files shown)


OwenCorrigan76:
Oh no. I'll have a look there now. Thanks

ppatierno (approved):

PR #97, opened 2023-01-18
Refactor Client-Examples ProducerConfigs 

OwenCorrigan76:
Refactor  Producer Config files in the Java client by reducing the complexity and size of the class. Standardise the naming scheme of environment variables used to configure the clients / standardise the envVars used to configure the Kafka client properties as outlined here [1].

Example of standardised Environmental Variables:

KAFKA_<NAME_OF_KAFKA_PROPERTY_DELIMITED_BY_UNDERSCORES>
[1] https://github.com/kyguy/proposals/blob/refactor-client-examples/040-refactor-client-examples.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,20 +59,13 @@ To run the OAuth example, you will need to have your Kafka cluster configured wi
 Below are listed and described the available environment variables that can be used for configuration.
 
 Producer  
-* `BOOTSTRAP_SERVERS` - comma-separated host and port pairs that is a list of Kafka broker addresses. The form of pair is `host:port`, e.g. `my-cluster-kafka-bootstrap:9092` 
-* `TOPIC` - the topic the producer will send to  
-* `DELAY_MS` - the delay, in ms, between messages  
-* `MESSAGE_COUNT` - the number of messages the producer should send  
-* `CA_CRT` - the certificate of the CA which signed the brokers' TLS certificates, for adding to the client's trust store
-* `USER_CRT` - the user's certificate
-* `USER_KEY` - the user's private key
-* `LOG_LEVEL` - logging level  
-* `PRODUCER_ACKS` - acknowledgement level
-* `HEADERS` - custom headers list separated by commas of `key1=value1, key2=value2`
-* `BLOCKING_PRODUCER` - if it's set, the producer will block another message until ack will be received
-* `MESSAGES_PER_TRANSACTION` - how many messages will be part of one transaction. Transaction config could be set via `ADDITIONAL_CONFIG` variable. Default is 10.
-* `ADDITIONAL_CONFIG` - additional configuration for a producer application. Notice, that you can also override any previously set variable by setting this. The form is `key=value` records separated by new line character
-* `TRACING_SYSTEM` - if it's set to `jaeger` or `opentelemetry`, this will enable tracing. 
+* `STRIMZI_TOPIC` - the topic the producer will send to  
+* `STRIMZI_DELAY_MS` - the delay, in ms, between messages  
+* `STRIMZI_MESSAGE_COUNT` - the number of messages the producer should send
+* `STRIMZI_MESSAGE` - the message the producer will send
+* `STRIMZI_LOG_LEVEL` - logging level  
+* `STRIMZI_HEADERS` - custom headers list separated by commas of `key1=value1, key2=value2`
+* `STRIMZI_TRACING_SYSTEM` - if it's set to `jaeger` or `opentelemetry`, this will enable tracing. 
 
 Consumer  
 * `STRIMZI_TOPIC` - name of topic which consumer subscribes  
@@ -80,7 +73,7 @@ Consumer
 * `STRIMZI_LOG_LEVEL` - logging level  
 * `STRIMZI_TRACING_SYSTEM` - if it's set to `jaeger` or `opentelemetry`, this will enable tracing.
 
-Additionally, any Kafka Consumer API configuration option can be passed as an environment variable.
+Additionally, any Kafka Consumer API or Kafka Producer API configuration option can be passed as an environment variable.
 It should be prefixed with `KAFKA_` and use `_` instead of `.`.
 For example environment variable `KAFKA_BOOTSTRAP_SERVERS` will be used as the `bootstrap.servers` configuration option in the Kafka Consumer API.
 
--- a/java/kafka/deployment-oauth.yaml
+++ b/java/kafka/deployment-oauth.yaml
@@ -69,16 +69,20 @@ spec:
       - name: java-kafka-producer
         image: quay.io/strimzi-examples/java-kafka-producer:latest
         env:
-          - name: BOOTSTRAP_SERVERS
+          - name: KAFKA_BOOTSTRAP_SERVERS
             value: my-cluster-kafka-bootstrap:9092
-          - name: TOPIC
+          - name: STRIMZI_TOPIC
             value: my-topic
-          - name: DELAY_MS
+          - name: STRIMZI_DELAY_MS
             value: "1000"
-          - name: LOG_LEVEL
+          - name: STRIMZI_LOG_LEVEL
             value: "INFO"
-          - name: MESSAGE_COUNT
+          - name: STRIMZI_MESSAGE_COUNT
             value: "1000000"
+          - name: KAFKA_KEY_SERIALIZER
+            value: "org.apache.kafka.common.serialization.StringSerializer"
+          - name: KAFKA_VALUE_SERIALIZER
+            value: "org.apache.kafka.common.serialization.StringSerializer"
           - name: OAUTH_CLIENT_ID
             value: java-kafka-producer
           - name: OAUTH_CLIENT_SECRET
--- a/java/kafka/deployment-oauth.yaml
+++ b/java/kafka/deployment-oauth.yaml
@@ -18,29 +18,6 @@ spec:
   replicas: 3
   partitions: 12
 ---
-apiVersion: kafka.strimzi.io/v1beta1
-kind: KafkaUser
-metadata:
-  name: java-kafka-producer
-  labels:
-    strimzi.io/cluster: my-cluster
-spec:
-  authorization:
-    type: simple
-    acls:
-      - resource:
-          type: topic
-          name: my-topic
-        operation: Write
-      - resource:
-          type: topic
-          name: my-topic
-        operation: Create
-      - resource:
-          type: topic
-          name: my-topic
-        operation: Describe
----
 apiVersion: v1
 kind: Secret
 metadata:
--- a/java/kafka/deployment-oauth.yaml
+++ b/java/kafka/deployment-oauth.yaml
@@ -25,8 +25,6 @@ metadata:
   labels:
     strimzi.io/cluster: my-cluster
 spec:
-  authentication:
-    type: tls
   authorization:
     type: simple
     acls:
@@ -71,8 +69,6 @@ spec:
         - name: java-kafka-producer
           image: quay.io/strimzi-examples/java-kafka-producer:latest
           env:
-            - name: KAFKA_BOOTSTRAP_SERVERS
-              value: my-cluster-kafka-bootstrap:9093
             - name: STRIMZI_TOPIC
               value: my-topic
             - name: STRIMZI_DELAY_MS
@@ -81,10 +77,6 @@ spec:
               value: "INFO"
             - name: STRIMZI_MESSAGE_COUNT
               value: "1000000"
-            - name: KAFKA_KEY_SERIALIZER
-              value: "org.apache.kafka.common.serialization.StringSerializer"
-            - name: KAFKA_VALUE_SERIALIZER
-              value: "org.apache.kafka.common.serialization.StringSerializer"
             - name: OAUTH_CLIENT_ID
               value: java-kafka-producer
             - name: OAUTH_CLIENT_SECRET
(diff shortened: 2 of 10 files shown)


kyguy:
Needs a rebase on `main` branch

kyguy (approved):

ppatierno:
@OwenCorrigan76 Have you double checked that the YAMLs now work ok on a Kubernetes cluster?

OwenCorrigan76:
@ppatierno Yes. I've tested all the yaml files, including spending yesterday and today working through Marko and Aykut's posts to get the oauth and ssl yaml's working. All yaml 's worked as they should and the producer was sending messages in the cluster.

ppatierno (approved):
LGTM

java/kafka/java-kafka-producer.yaml
```diff
@@ -18,18 +18,20 @@ spec:
...
             value: "1000"
-          - name: LOG_LEVEL
+          - name: STRIMZI_LOG_LEVEL
             value: "INFO"
-          - name: MESSAGE_COUNT
+          - name: STRIMZI_MESSAGE_COUNT
             value: "1000000"
-          - name: PRODUCER_ACKS
+          - name: STRIMZI_PRODUCER_ACKS
             value: "all"
```
scholzj: This does not seem to be mentioned anywhere else. Should it be `KAFKA_ACKS`?

java/kafka/producer/src/main/java/KafkaProducerExample.java
```diff
@@ -58,7 +60,7 @@ public static void main(String[] args) throws InterruptedException {
         log.info("Sending {} messages ...", config.getMessageCount());
 
         boolean blockProducer = System.getenv("BLOCKING_PRODUCER") != null;
-        boolean transactionalProducer = config.getAdditionalConfig().contains("transactional.id");
+        boolean transactionalProducer = Boolean.parseBoolean(props.getProperty(TRANSACTIONAL_ID_CONFIG));
```
scholzj: `transactional.id` is a String config option. I do not think parsing it to boolean makes any sense. The original logic should be probably used.

README.md
```diff
@@ -59,28 +59,21 @@ To run the OAuth example, you will need to have your Kafka cluster configured wi
...
-* `MESSAGES_PER_TRANSACTION` - how many messages will be part of one transaction. Transaction config could be set via `ADDITIONAL_CONFIG` variable. Default is 10.
-* `ADDITIONAL_CONFIG` - additional configuration for a producer application. Notice, that you can also override any previously set variable by setting this. The form is `key=value` records separated by new line character
-* `TRACING_SYSTEM` - if it's set to `jaeger` or `opentelemetry`, this will enable tracing. 
+* `STRIMZI_TOPIC` - the topic the producer will send to  
+* `STRIMZI_DELAY_MS` - the delay, in ms, between messages  
+* `STRIMZI_MESSAGE_COUNT` - the number of messages the producer should send
+* `STRIMZI_MESSAGE` - the message the producer will send
+* `STRIMZI_LOG_LEVEL` - logging level  
+* `STRIMZI_HEADERS` - custom headers list separated by commas of `key1=value1, key2=value2`
+* `STRIMZI_TRACING_SYSTEM` - if it's set to `jaeger` or `opentelemetry`, this will enable tracing.
```
scholzj, line 68: You talk about `STRIMZI_TRACING_SYSTEM`, but you do not seem to use it in the producer or in the previous PR in the consumer.

java/kafka/deployment-oauth.yaml
```diff
@@ -69,16 +69,20 @@ spec:
...
-          - name: LOG_LEVEL
+          - name: STRIMZI_LOG_LEVEL
             value: "INFO"
-          - name: MESSAGE_COUNT
+          - name: STRIMZI_MESSAGE_COUNT
             value: "1000000"
+          - name: KAFKA_KEY_SERIALIZER
+            value: "org.apache.kafka.common.serialization.StringSerializer"
+          - name: KAFKA_VALUE_SERIALIZER
+            value: "org.apache.kafka.common.serialization.StringSerializer"
```
scholzj: Is this really enough? Did you tried it? My guess is that both here (but also in the consumer), you are missing at least:
* Security protocol which needs to be set to PLAINTEXT_SASL or something like that
* SASL mechanism configuration
* The JAAS configuration
* Callback handler configuration

java/kafka/deployment-oauth.yaml
```diff
@@ -66,35 +68,54 @@ spec:
...
+            - name: KAFKA_KEY_SERIALIZER
+              value: "org.apache.kafka.common.serialization.StringSerializer"
+            - name: KAFKA_VALUE_SERIALIZER
+              value: "org.apache.kafka.common.serialization.StringSerializer"
+            - name: OAUTH_CLIENT_ID
+              value: java-kafka-producer
+            - name: OAUTH_CLIENT_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: java-kafka-producer
```
kyguy: To match the name of the secret with the token
```suggestion
                  name: java-kafka-producer-oauth
```

java/kafka/deployment-oauth.yaml
```diff
@@ -66,35 +68,54 @@ spec:
...
-          - name: OAUTH_SSL_TRUSTSTORE_CERTIFICATES
-            valueFrom:
-              secretKeyRef:
-                name: sso-x509-https-secret
-                key: tls.crt
+        - name: java-kafka-producer
+          image: quay.io/strimzi-examples/java-kafka-producer:latest
+          env:
+            - name: KAFKA_BOOTSTRAP_SERVERS
+              value: my-cluster-kafka-bootstrap:9093
```
kyguy: If we configure our Kafka cluster to use OAuth plain like in this example [1] we can then continue using the insecure port at 9092, remove `KAFKA_SSL_TRUSTSTORE_TYPE`, `KAFKA_SSL_TRUSTSTORE_CERTIFICATES` env var fields and even  `authentication` section from the java-kafka-producer `KafkaUser`

I know we tested this with a Kafka cluster with TLS authentication enabled, if we instead test with a plain listener like this [1] we could avoid having to adding these things to the example!

[1] https://github.com/strimzi/strimzi-kafka-oauth/blob/main/examples/kubernetes/kafka-oauth-over-plain-single-authz.yaml

java/kafka/producer/src/main/java/KafkaProducerExample.java
kyguy: @OwenCorrigan76 it looks like the fix for this got lost in the shuffle, we still need to commit the fix you made for this here

java/kafka/java-kafka-producer.yaml
OwenCorrigan76: Yes it should. Changed to KAFKA_ACKS

java/kafka/producer/src/main/java/KafkaProducerExample.java
OwenCorrigan76: Logic Fixed

README.md
OwenCorrigan76: STRIMZI prefix added where needed

java/kafka/deployment-oauth.yaml
kyguy: Testing with a broker listener configuration:
```
kind: Kafka
...
spec:
  kafka:
    ...
    listeners:
    - name: plain
      port: 9092
      tls: false
      type: internal
    - authentication:
        type: oauth
        enablePlain: true
        userNameClaim: preferred_username
        jwksEndpointUri: http://keycloak:8080/auth/realms/kafka-authz/protocol/openid-connect/certs
        validIssuerUri: http://keycloak:8080/auth/realms/kafka-authz
        maxSecondsWithoutReauthentication: 3600
```
Allows us to strip out:
- `KAFKA_SSL_TRUSTSTORE_TYPE`
- `KAFKA_SSL_TRUSTSTORE_CERTIFICATES`
- `OAUTH_SSL_TRUSTSTORE_CERTIFICATES`
- `OAUTH_SSL_TRUSTSTORE_TYPE`

However, we need to add in:
```
name: KAFKA_SECURITY_PROTOCOL
value: "SASL_PLAINTEXT"
```

scholzj: OAuth without encryption would be insecure. So you should use TLS encryption here IMHO.

kyguy: Ignore my comments here @OwenCorrigan76! It is great as it is

java/kafka/deployment-oauth.yaml
```diff
@@ -21,10 +21,12 @@ spec:
 apiVersion: kafka.strimzi.io/v1beta1
 kind: KafkaUser
 metadata:
-  name: service-account-java-kafka-producer
+  name: java-kafka-producer
   labels:
     strimzi.io/cluster: my-cluster
 spec:
+  authentication:
+    type: tls
```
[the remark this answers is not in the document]
kyguy: Ignore this comment

java/kafka/deployment-oauth.yaml
```diff
@@ -66,35 +68,54 @@ spec:
...
+              value: OAUTHBEARER
+            - name: KAFKA_SASL_LOGIN_CALLBACK_HANDLER_CLASS
+              value: "io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler"
+            - name: KAFKA_SSL_TRUSTSTORE_TYPE
+              value: PEM
+            - name: KAFKA_SSL_TRUSTSTORE_CERTIFICATES
+              valueFrom:
+                secretKeyRef:
+                  name: my-cluster-cluster-ca-cert
+                  key: ca.crt
```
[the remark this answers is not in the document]
kyguy: Ignore

kyguy (approved):
Looks great @OwenCorrigan76!

java/kafka/deployment-oauth.yaml
OwenCorrigan76: Added and tested all of the above including testing both SSL and  SASL_PLAINTEXT for Security Protocol.

java/kafka/deployment-oauth.yaml
scholzj: Why do we use the KafkaUser here? With OAuth, the user would be managed in the OAuth server. So why do we use it?

Same applies to the consumer I guess?

scholzj (commented):
I left one more comment. Also, a general point ... maybe you can think about some ordering of the environment variables in the Deployment files? E.g. first all `STRIMZI_`, then all `OAUTH_`, then all `KAFKA_`? Seems a bit random right now, so it is hard to get a quick overview of what is used where.

java/kafka/deployment-oauth.yaml
OwenCorrigan76: Oversight. I forgot to remove it. Removed now.

OwenCorrigan76: @scholzj I've re-ordered the envVars as recommended. I've also realigned the Producer yaml's so they are the same as Streams and Consumer.

scholzj: Just to be clear ... I would probably drop the KafkaUser completely ... I'm not sure I understand its value here.

scholzj: I mean in the OAuth examples only of course.

OwenCorrigan76: KafkaUser removed from Producer

java/kafka/producer/src/main/java/KafkaProducerExample.java
```diff
@@ -55,10 +55,9 @@ public static void main(String[] args) throws InterruptedException {
         }
 
         KafkaProducer<String, String> producer = new KafkaProducer<>(props);
-        log.info("Sending {} messages ...", config.getMessageCount());
 
         boolean blockProducer = System.getenv("BLOCKING_PRODUCER") != null;
-        boolean transactionalProducer = config.getAdditionalConfig().contains("transactional.id");
+        boolean transactionalProducer = System.getenv("KAFKA_TRANSACTIONAL_ID") != null;
         int msgPerTx = Integer.parseInt(System.getenv().getOrDefault("MESSAGES_PER_TRANSACTION", "10"));
```
scholzj: Same as above, should this be removed or `STRIMZI_MESSAGES_PER_TRANSACTION`? (plus documented in the README?)

java/kafka/producer/src/main/java/KafkaProducerExample.java
```diff
@@ -55,10 +55,9 @@ public static void main(String[] args) throws InterruptedException {
         }
 
         KafkaProducer<String, String> producer = new KafkaProducer<>(props);
-        log.info("Sending {} messages ...", config.getMessageCount());
 
         boolean blockProducer = System.getenv("BLOCKING_PRODUCER") != null;
```
scholzj: Should this be a `STRIMZI_BLOCKING_PRODUCER`? Or should we remove it?

scholzj (commented):
There seem to be some left-over configs for the blocking and transaction handling. I guess you should either update them to use the right naming and document them. Or maybe simply remove the whole transactional and blocking logic as not important for the exmples? I guess those were mainly from when it was used for tests?

java/kafka/producer/src/main/java/KafkaProducerExample.java
OwenCorrigan76: @scholzj  Logic removed

scholzj (approved):

scholzj:
Thanks for the PR.

PR #105, opened 2023-02-18
Add Java SDK HTTP client examples

ppatierno:
This PR provides client examples using the Java SDK HTTP client to interact with a Strimzi HTTP bridge in order to send and receive messages.
The tracing support is just for OpenTelemetry, taking into account that OpenTracing is now deprecated (in all the Strimzi ecosystem) and will be definitively removed soon.


java/http/java-http-consumer/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+
+USER root
+
+RUN microdnf update \
+    && microdnf install java-17-openjdk-headless shadow-utils \
```
scholzj: You should install the Java as we do here to get rid of some CVEs in the images. Both here as well as in the other Dockerfile: https://github.com/strimzi/strimzi-kafka-operator/blob/main/docker-images/base/Dockerfile#L11

scholzj (commented):
Did you create the new images on Quay and made sure the CI has the rights to push them?

ppatierno:
I didn't take a look at the CI but you are right.
I thought that it was going to build and push images when the PR was merged.
Just to understand, should I build images locally and push them to the Strimzi quay.io organization directly?

java/http/java-http-consumer/Dockerfile
ppatierno: Ok I will change this way but it seems we need to do the same for the other examples, referring to the Java Kafka ones. There is the same problem there as well. But not for this PR.

scholzj:
> I didn't take a look at the CI but you are right.
> I thought that it was going to build and push images when the PR was merged.
> Just to understand, should I build images locally and push them to the Strimzi quay.io organization directly?

If the CI has the rights, it will create the repositories as private and you would need to make them public. But the CI might not have the right to do it in the first place as it is usually set-up to be able to push only to the repositories it needs.

ppatierno:
Right. Going to build and push them.

ppatierno:
@scholzj I have built images locally and pushed under the Quay.io strimzi-examples organization making them public.

scholzj (approved):

PR #109, opened 2023-02-20
Replace OpenTelemetry Jaeger exporter with the default OTLP exporter

### LINKED ISSUE #102: Replace OpenTelemetry Jaeger exporter with the default OTLP exporter
Strimzi components already moved to use the OTLP exporter (instead of the Jaeger one) for the OpenTelemetry support.
I think that the examples should do the same, as it is the default.
In order to do that, examples should not have the `opentelemetry-exporter-jaeger` dependency but the `opentelemetry-exporter-otlp`.
We can even get rid of the `OTEL_TRACES_EXPORTER` env var in the YAMLs because OTLP is the default protocol for OpenTelemetry. We should also replace `OTEL_EXPORTER_JAEGER_ENDPOINT` with `OTEL_EXPORTER_OTLP_ENDPOINT` env var (it uses port 4317).
Finally, this should be anyway tested with a Jaeger backend exposing an OTLP endpoint. This can be enabled starting up Jaeger with `COLLECTOR_OTLP_ENABLED=true` env var.

OwenCorrigan76:
This PR fixes #102
Strimzi components have already moved to use the OTLP exporter (instead of the Jaeger one) for the OpenTelemetry support.
The client-examples should do the same, as it is the default.
Replace opentelemetry-exporter-jaeger dependency with the opentelemetry-exporter-otlp dependency.
Remove OTEL_TRACES_EXPORTER env var in the YAMLs because OTLP is the default protocol for OpenTelemetry. 
Replace OTEL_EXPORTER_JAEGER_ENDPOINT with OTEL_EXPORTER_OTLP_ENDPOINT env var, using port 4317.

### DIFF
--- a/java/kafka/deployment-tracing-opentelemetry.yaml
+++ b/java/kafka/deployment-tracing-opentelemetry.yaml
@@ -56,12 +56,10 @@ spec:
               value: "org.apache.kafka.common.serialization.StringSerializer"
             - name: OTEL_SERVICE_NAME
               value: kafka-otel
-            - name: OTEL_TRACES_EXPORTER
-              value: jaeger
             - name: OTEL_METRICS_EXPORTER
               value: none
-            - name: OTEL_EXPORTER_JAEGER_ENDPOINT
-              value: http://my-jaeger-collector:14250
+            - name: OTEL_EXPORTER_OTLP_ENDPOINT
+              value: http://my-jaeger-collector:4317
 ---
 apiVersion: apps/v1
 kind: Deployment
@@ -103,12 +101,10 @@ spec:
               value: "org.apache.kafka.common.serialization.Serdes$StringSerde"
             - name: OTEL_SERVICE_NAME
               value: kafka-otel
-            - name: OTEL_TRACES_EXPORTER
-              value: jaeger
             - name: OTEL_METRICS_EXPORTER
               value: none
-            - name: OTEL_EXPORTER_JAEGER_ENDPOINT
-              value: http://my-jaeger-collector:14250
+            - name: OTEL_EXPORTER_OTLP_ENDPOINT
+              value: http://my-jaeger-collector:4317
 ---
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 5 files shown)


java/kafka/deployment-tracing-opentelemetry.yaml
```diff
@@ -56,12 +56,10 @@ spec:
...
             - name: OTEL_SERVICE_NAME
               value: kafka-otel
-            - name: OTEL_TRACES_EXPORTER
-              value: jaeger
             - name: OTEL_METRICS_EXPORTER
               value: none
-            - name: OTEL_EXPORTER_JAEGER_ENDPOINT
-              value: http://my-jaeger-collector:14250
+            - name: OTEL_EXPORTER_OTLP_ENDPOINT
+              value: http://my-jaeger-collector:4317
```
ppatierno: I would use `http://my-otlp-endpoint:4317` everywhere. It makes clear that it's an OTLP endpoint needed, not Jaeger (even if Jaeger backend exposes anyway an OTLP endpoint).

ppatierno (changes requested):
I would assume you tried these change on Kubernetes running Jaeger exposing an OTLP endpoint, right?

java/kafka/deployment-tracing-opentelemetry.yaml
OwenCorrigan76: @ppatierno Changed to `http://my-otlp-endpoint:4317`.
It was tested and working.

ppatierno (approved):

im-konge (approved):

OwenCorrigan76:
@ppatierno Is this ready for merge? Thanks

ppatierno:
@OwenCorrigan76 I am fine merging, but didn't get any answer to my question if it was tested properly. Can you enlight me please?

java/kafka/deployment-tracing-opentelemetry.yaml
OwenCorrigan76: @ppatierno See above from Monday. Thanks

ppatierno: Great, I missed this comment! Thanks!

PR #113, opened 2023-03-01
Move convertEnvToPropertyKey method to Common Module

OwenCorrigan76:
To further reduce the verbosity in the Java-Kafka Client-Examples, any common occurring methods in Kafka consumer, producer or streams directories will be moved to the ConfigUtils class in the Common module, starting here with the convertEnvToPropertyKey method.

### DIFF
--- a/java/common/src/main/java/ConfigUtil.java
+++ b/java/common/src/main/java/ConfigUtil.java
@@ -3,7 +3,7 @@
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
 
-public class ConvertEnvToPropKey {
+public class ConfigUtil {
 
     public static String convertEnvVarToPropertyKey(String envVar) {
         return envVar.substring(envVar.indexOf("_") + 1).toLowerCase().replace("_", ".");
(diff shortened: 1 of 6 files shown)


java/common/src/main/java/ConfigUtil.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
```
scholzj: This should be part of some package.

OwenCorrigan76: Package created in common module. Will add subsequent packages shortly.

scholzj (approved):

java/common/src/main/java/io/strimzi/common/ConfigUtil.java
```diff
@@ -0,0 +1,13 @@
...
+ */
+package io.strimzi.common;
+
+public class ConfigUtil {
+
+    public static String convertEnvVarToPropertyKey(String envVar) {
+        return envVar.substring(envVar.indexOf("_") + 1).toLowerCase().replace("_", ".");
+    }
+
+}
```
ppatierno, line 20: I would just add a little bit of Javadoc both on class and method to avoid Java 17 support complaining in the future.

OwenCorrigan76: @ppatierno javadoc added to ConfigUtil class in Common module.

ppatierno (approved):

scholzj:
Thanks for the PR

PR #114, opened 2023-03-06
Create a Python client in Client-Examples

OwenCorrigan76:
The Strimzi Client-Examples repository had support for Java, Golang and HTTP Vertex, but there was no support for Python. Since Python is a powerful and widely used language, it would be a valuable addition to the current Strimzi portfolio to have this language supported.


python/kafka/consumer/kafka_consumer.py
```diff
@@ -0,0 +1,17 @@
+from kafka import KafkaConsumer
+from datetime import datetime
+from kafka_consumer_config import from_env
+import logging
+
+if __name__ == '__main__':
+
+    config = from_env()
+    logging.info(f"Go consumer starting with config={config}")
```
scholzj:
```suggestion
    logging.info(f"Python consumer starting with config={config}")
```

python/kafka/producer/kafka_producer.py
```diff
@@ -0,0 +1,21 @@
+from kafka import KafkaProducer
+from datetime import datetime
+from kafka_producer_config import from_env
+import time
+
+if __name__ == '__main__':
+    config = from_env()
+    print(config.properties)
```
mimaison: In the consumer we used `logging` instead of `print`. Let's be consistent

python/kafka/consumer/kafka_consumer_config.py
```diff
@@ -0,0 +1,48 @@
...
+DEFAULT_BOOTSTRAP_SERVERS_ = "localhost:9092"
+DEFAULT_TOPIC = "my-topic"
+DEFAULT_GROUP_ID = "my-group"
+DEFAULT_MESSAGE_COUNT = 10
+
+# default values for environmental variables in a dictionary
+DEFAULT_PROPERTIES = {
+    'bootstrap_servers': 'localhost:9092',
+    'group_id': 'my-java-kafka-consumer',
+    'client_id': '',
```
mimaison: Should we put a default client id?

python/kafka/consumer/kafka_consumer_config.py
```diff
@@ -0,0 +1,48 @@
...
+    topic = get_env(TOPIC, DEFAULT_TOPIC)
+    group_id = get_env(GROUP_ID, DEFAULT_GROUP_ID)
+    message_count = get_env(MESSAGE_COUNT, DEFAULT_MESSAGE_COUNT)
+    for k, v in os.environ.items():
+        if 'KAFKA_' in k:
+            DEFAULT_PROPERTIES[convertEnvVarToPropertyKey(k)] = v
+    return KafkaConsumerConfig(topic, group_id, message_count, DEFAULT_PROPERTIES)
+
+
+def get_env(env_var, default_value):
```
mimaison: Do we really need this method? All it does is replace `os.getenv` by `get_env`

python/kafka/producer/kafka_producer_config.py
```diff
@@ -0,0 +1,57 @@
...
+    message_count = get_env(MESSAGE_COUNT, DEFAULT_MESSAGE_COUNT)
+    properties = DEFAULT_PROPERTIES
+
+    for k, v in os.environ.items():
+        if 'KAFKA_' in k:
+            properties[convertEnvVarToPropertyKey(k)] = v
+    return KafkaProducerConfig(topic, message, group_id, delay_ms, message_count, properties)
+
+
+def get_env(env_var, default_value):
```
mimaison: Do we really need this method? All it does is replace `os.getenv` by `get_env`

python/kafka/producer/kafka_producer_config.py
```diff
@@ -0,0 +1,57 @@
...
+        self.delay = delay_ms
+        self.properties = properties
+
+
+def from_env():
+    topic = get_env(TOPIC, DEFAULT_TOPIC)
+    group_id = get_env(GROUP_ID, DEFAULT_GROUP_ID)
+    message = get_env(MESSAGE, DEFAULT_MESSAGE)
+    delay_ms = get_env(DELAY, DEFAULT_DELAY)
+    message_count = get_env(MESSAGE_COUNT, DEFAULT_MESSAGE_COUNT)
```
mimaison: It seems that `message_count` and `delay` are not used

python/kafka/producer/kafka_producer_config.py
```diff
@@ -0,0 +1,57 @@
...
+# convert the kafka (strimzi) prefix
+def convertEnvVarToPropertyKey(e):
+    return e[e.index('_') + 1:].lower()
+
+
+class KafkaProducerConfig:
+    def __init__(self, topic, message, group_id, message_count, delay_ms, properties):
+        self.topic = topic
+        self.message = message
+        self.group_id = group_id
```
mimaison: Should this be `client_id` instead?

mimaison (commented):
Thanks for the PR.

I left a few comments. Can we also remove the `pyc` files and include a [requirements file](https://pip.pypa.io/en/stable/user_guide/#requirements-files) so it's clear which dependencies are being used.

python/kafka/consumer/kafka_consumer_config.py
OwenCorrigan76: Removed

python/kafka/consumer/kafka_consumer_config.py
OwenCorrigan76: Method removed

python/producer/__pycache__/kafka_producer_config.cpython-310.pyc
scholzj, line 1: I guess this should not be committed?

python/consumer/__pycache__/kafka_consumer_config.cpython-310.pyc
scholzj, line 1: I guess this should not be committed?

java/common/src/main/java/io/strimzi/common/ConfigUtil.java
scholzj, line 1: This probably should not be here?

python/Makefile.docker
scholzj, line 1: Should the commented out parts be deleted? Also, how does it differ from the existing Makefile for building the other images?

python/consumer/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+
+USER root
+
+RUN microdnf update \
+    && microdnf install python3-devel shadow-utils \
+    && microdnf clean all
+
+
+
```
scholzj: One empty line would be enough.

python/consumer/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# Add strimzi user with UID 1001
+# The user is in the group 0 to have access to the mounted volumes and storage
+RUN useradd -r -m -u 1001 -g 0 strimzi
+
+
```
scholzj: Same as above.

python/consumer/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+
+USER root
+
+RUN microdnf update \
+    && microdnf install python3-devel shadow-utils \
```
scholzj: I have inimal experience with Python ... is this a common way to build a Python application image to install `python-devel` and copy the dependencies with `pip install` inside the image?

scholzj (commented):
Just some quick things ...

python/kafka/producer/kafka_producer.py
OwenCorrigan76: print used throughout

python/kafka/producer/kafka_producer_config.py
OwenCorrigan76: group_id removed

python/kafka/consumer/kafka_consumer.py
OwenCorrigan76: removed

python/producer/__pycache__/kafka_producer_config.cpython-310.pyc
OwenCorrigan76: removed

python/consumer/__pycache__/kafka_consumer_config.cpython-310.pyc
OwenCorrigan76: removed

java/common/src/main/java/io/strimzi/common/ConfigUtil.java
OwenCorrigan76: removed

python/consumer/Dockerfile
OwenCorrigan76: Empty lines removed

python/consumer/Dockerfile
OwenCorrigan76: Removed

python/consumer/Dockerfile
OwenCorrigan76: Looking into this

python/Makefile.docker
OwenCorrigan76: Only difference is that the scripts aren't being called in Python/Makefile.docker. They were commented out. I've deleted them but will return to this.

python/kafka/producer/kafka_producer_config.py
OwenCorrigan76: also removed

python/kafka/producer/kafka_producer_config.py
OwenCorrigan76: Not needed. Removed

python/consumer/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
```
mimaison: I wonder if we should start from a ubi-python images instead of installing python ourselves. WDYT?

python/consumer/Dockerfile
mimaison: I've seen it done this way in other projects, ie run pip install in the image

python/consumer/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    && microdnf clean all
+
+# Add strimzi user with UID 1001
+# The user is in the group 0 to have access to the mounted volumes and storage
+RUN useradd -r -m -u 1001 -g 0 strimzi
+
+ARG version=latest
+ENV VERSION ${version}
+
+ADD kafka_consumer.py /
```
mimaison: Should we have a WORKDIR instead of copying everything in `/`?

python/consumer/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ADD kafka_consumer_config.py /
+ADD requirements.txt /
+
+RUN pip3 install -r requirements.txt
+
+USER 1001
+
+CMD ["python3", "/kafka_consumer.py"]
```
mimaison: It's usually best practice to have a new line. Same in the other files

python/consumer/kafka_consumer_config.py
```diff
@@ -0,0 +1,39 @@
...
+MESSAGE_COUNT = "STRIMZI_MESSAGE_COUNT"
+
+# default values for Strimzi environment variables
+DEFAULT_TOPIC = "my-topic"
+DEFAULT_MESSAGE_COUNT = 10
+
+# default values for Kafka environmental variables in a dictionary
+DEFAULT_PROPERTIES = {
+    'bootstrap.servers': 'localhost:9092',
+    'group.id': 'my-kafka-consumer',
```
mimaison: Let's add a `client.id` too

python/consumer/kafka_consumer_config.py
```diff
@@ -0,0 +1,39 @@
...
+    'auto.offset.reset': 'latest',
```
mimaison: This is the default value, so maybe we can remove it?

python/producer/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+
+# Add strimzi user with UID 1001
+# The user is in the group 0 to have access to the mounted volumes and storage
+RUN useradd -r -m -u 1001 -g 0 strimzi
+
+ARG version=latest
+ENV VERSION ${version}
+
+#COPY ./scripts/ /bin
```
mimaison: Is this needed?

python/consumer/kafka_consumer_config.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def convertEnvVarToPropertyKey(e):
+    return e[e.index('_') + 1:].lower()
+
+
+class KafkaConsumerConfig:
+    def __init__(self, topic, message_count, properties):
+        self.topic = topic
+        self.message_count = message_count
```
mimaison: It looks like we don't use this anywhere

python/producer/kafka_producer_config.py
```diff
@@ -0,0 +1,41 @@
...
+KAFKA_PREFIX = "KAFKA_"
+STRIMZI_PREFIX = "STRIMZI"
+
+# Strimzi environment variables declaration
+TOPIC = "STRIMZI_TOPIC"
+MESSAGE = "STRIMZI_MESSAGE"
+
+# default values for Strimzi environment variables
+DEFAULT_TOPIC = "my-topic"
+DEFAULT_MESSAGE = 'Hello to the World'
```
mimaison: We have a mix of single and double quotes. Can we stick to one type?

python/requirements.txt
```diff
@@ -0,0 +1 @@
+confluent_kafka==1.5.0
```
mimaison: Do we need this file?

mimaison (commented):
Thanks for the updates. I left a few more comments. Can we also have a README explaining how to run this sample?

python/producer/Dockerfile
OwenCorrigan76: Removed

python/producer/kafka_producer_config.py
OwenCorrigan76: Done

python/consumer/kafka_consumer_config.py
OwenCorrigan76: I had planned on introducing some more logic like ` while receivedMsgs < config.message_count: .....` to the consumer but for now I'll keep it simple and remove message_count

python/consumer/kafka_consumer_config.py
OwenCorrigan76: Removed

python/requirements.txt
OwenCorrigan76: No. Removed

OwenCorrigan76: no- removed

python/consumer/Dockerfile
OwenCorrigan76: Added

python/consumer/Dockerfile
OwenCorrigan76: I've switched to use `/ubi8/python-38:latest` as recommended by Mickael, and that removed the need to run `microdnf install python3-devel` inside the image.

python/consumer/Dockerfile
OwenCorrigan76: I've switched to uni-python and removed `RUN microdnf update \
    && microdnf install python3-devel shadow-utils \
    && microdnf clean all`

python/consumer/Dockerfile
OwenCorrigan76: I was following the format of the other Dockerfiles in the Client Examples. There's no WORKDIR. I did add HOME directory so not everything is being copied. I can add a WORKDIR if you think it is better.

`ADD kafka_consumer.py $HOME` etc.

OwenCorrigan76 (commented):
Fix README re deployment.yaml

OwenCorrigan76 (commented):
Fix CMD directory in Dockerfiles

OwenCorrigan76 (commented):
Change `python-deployment.yaml` to `python-kafka-deployment.yaml` and change `env:` `name:` `KAFKA_BOOTSTRAP_SERVERS` to `KAFKA_BOOTSTRAP.SERVERS`

python/README.md
```diff
@@ -0,0 +1,62 @@
+[![Build Status](https://dev.azure.com/cncf/strimzi/_apis/build/status/client-examples?branchName=main)](https://dev.azure.com/cncf/strimzi/_build/latest?definitionId=33&branchName=main)
+[![License](https://img.shields.io/badge/license-Apache--2.0-blue.svg)](http://www.apache.org/licenses/LICENSE-2.0)
+[![Twitter Follow](https://img.shields.io/twitter/follow/strimziio?style=social)](https://twitter.com/strimziio)
```
kyguy: We can remove these from this file

python/consumer/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM registry.access.redhat.com/ubi8/python-39:latest
+
+USER root
+
+# Add strimzi user with UID 1002
+# The user is in the group 0 to have access to the mounted volumes and storage
+RUN useradd -r -m -u 1002 -g 0 strimzi
```
kyguy: I know we switched the uid from 1001 (which is used in the Java example Dockerfiles) to 1002 to get around build issues. Not a huge priority but is there a way we can update this shell command to add uid 1001 to group 0?

kyguy: On that note, do any of the examples use mounted volumes and storage? (Do even need this line?)

python/consumer/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ARG version=latest
+ENV VERSION ${version}
+
+ADD kafka_consumer.py $HOME
+ADD kafka_consumer_config.py $HOME
+ADD requirements.txt $HOME
+
+RUN pip3 install -r requirements.txt
+
+USER 1002
```
kyguy: If above comment is sorted we can switch this back to 1001 to match the Java Dockerfile

python/consumer/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+CMD ["python3", "/opt/app-root/src/kafka_consumer.py"]
```
kyguy:
```suggestion
CMD ["python3", "${HOME}kafka_consumer.py"]
```
Would something like this work?

python/consumer/kafka_consumer.py
```diff
@@ -0,0 +1,18 @@
...
+        config.properties
+    )
+    consumer.subscribe([config.topic])
+
+    while True:
+        message = consumer.poll(timeout=1.0)
+        if message is None:
+            print("Waiting...")
+        else:
+            print(f'Consuming message @ {datetime.now()}', (message.value()))
```
kyguy: I assume we rely on the delay from the poll() function as a delay/sleep for this loop right?

python/consumer/kafka_consumer_config.py
```diff
@@ -0,0 +1,35 @@
+import os
+
+KAFKA_PREFIX = 'KAFKA_'
```
kyguy: This isn't used

python/consumer/kafka_consumer_config.py
```diff
@@ -0,0 +1,35 @@
...
+class KafkaConsumerConfig:
+    def __init__(self, topic, properties):
+        self.topic = topic
+        self.properties = properties
+
+
+def from_env():
+    topic = os.getenv(TOPIC, DEFAULT_TOPIC)
+    for k, v in os.environ.items():
+        if k.startswith('KAFKA_'):
```
kyguy: Use the `KAFKA_PREFIX` variable above

python/consumer/kafka_consumer_config.py
```diff
@@ -0,0 +1,35 @@
...
+TOPIC = 'STRIMZI_TOPIC'
+
+# default values for Strimzi environment variables
+DEFAULT_TOPIC = 'my-topic'
+
+# default values for Kafka environmental variables in a dictionary
+DEFAULT_PROPERTIES = {
+    'bootstrap.servers': 'localhost:9092',
+    'group.id': 'my-kafka-consumer',
+    'client.id': ''
```
kyguy: Does Kafka provide a a default value for client.id? If so, what is used?

Also, do we set a default client.id for any of the other client examples?

python/consumer/kafka_consumer_config.py
```diff
@@ -0,0 +1,35 @@
+import os
+
+KAFKA_PREFIX = 'KAFKA_'
+STRIMZI_PREFIX = 'STRIMZI'
```
kyguy: This isn't used

python/producer/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM registry.access.redhat.com/ubi8/python-39:latest
+
+USER root
+
+# Add strimzi user with UID 1002 The user is in the group 0 to have access to the mounted volumes and storage
+RUN useradd -r -m -u 1002 -g 0 strimzi
```
kyguy: Same comment/question as above

python/producer/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+ARG version=latest
+ENV VERSION ${version}
+
+ADD kafka_producer.py $HOME
+ADD kafka_producer_config.py $HOME
+ADD requirements.txt $HOME
+
+RUN pip3 install -r requirements.txt
+
+USER 1002
```
kyguy: same comment as above

python/producer/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+CMD ["python3", "/opt/app-root/src/kafka_producer.py"]
```
kyguy: Same comment as above

python/producer/kafka_producer.py
```diff
@@ -0,0 +1,18 @@
...
+    producer = Producer(
+        config.properties
+    )
+
+    while True:
+        producer_message = config.message.encode('utf-8')
+        producer.produce(config.topic, producer_message)
+        producer.flush()
+        print(f'Producing message @ {datetime.now()} | Message = {str(producer_message)}')
+        time.sleep(1)
```
kyguy: Was this sleep/delay configurable in the other client examples via the MESSAGE_DELAY? Or was that for something else?

python/producer/kafka_producer_config.py
```diff
@@ -0,0 +1,42 @@
+import os
+
+KAFKA_PREFIX = 'KAFKA_'
+STRIMZI_PREFIX = 'STRIMZI'
```
kyguy: These two vars are not used

python/producer/kafka_producer_config.py
```diff
@@ -0,0 +1,42 @@
...
+        self.properties = properties
+
+
+def from_env():
+    topic = os.getenv(TOPIC, DEFAULT_TOPIC)
+    message = os.getenv(MESSAGE, DEFAULT_MESSAGE)
+    properties = DEFAULT_PROPERTIES
+
+    for k, v in os.environ.items():
+        if k.startswith('KAFKA_'):
```
kyguy: Use the `KAFKA_PREFIX` here instead

python/producer/kafka_producer_config.py
```diff
@@ -0,0 +1,42 @@
...
+DEFAULT_TOPIC = 'my-topic'
+DEFAULT_MESSAGE = 'Hello to the World'
+
+# default values for environmental variables in dictionary
+DEFAULT_PROPERTIES = {
+    'bootstrap.servers': 'localhost:9092'
+}
+
+
+# convert the kafka (strimzi) prefix
```
kyguy: Might be worth giving a more explicit description of what this function does e.g. Converts env var names to variable names that are compliant with Kafka properties file format

python/producer/requirements.txt
```diff
@@ -0,0 +1 @@
+confluent_kafka==2.0.2
```
kyguy: Looks like the build was fixed to use the latest library version right?

python/python-kafka-consumer.yaml
```diff
@@ -0,0 +1,24 @@
...
+    matchLabels:
+      app: python-kafka-consumer
+  template:
+    metadata:
+      labels:
+        app: python-kafka-consumer
+    spec:
+      containers:
+        - name: python-kafka-consumer
+          image: docker.io/owenerz/python-kafka-consumer:latest
```
kyguy, line 19: Don't forget to fix this image path to the strimzi repo in the quay registry

python/python-kafka-deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+    matchLabels:
+      app: python-kafka-producer
+  template:
+    metadata:
+      labels:
+        app: python-kafka-producer
+    spec:
+      containers:
+        - name: python-kafka-producer
+          image: docker.io/owenerz/python-kafka-producer:latest
```
kyguy, line 29: Same as above

python/python-kafka-deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+    matchLabels:
+      app: python-kafka-consumer
+  template:
+    metadata:
+      labels:
+        app: python-kafka-consumer
+    spec:
+      containers:
+        - name: python-kafka-consumer
+          image: docker.io/owenerz/python-kafka-consumer:latest
```
kyguy, line 56: Same as above

python/python-kafka-producer.yaml
```diff
@@ -0,0 +1,25 @@
...
+    matchLabels:
+      app: python-kafka-producer
+  template:
+    metadata:
+      labels:
+        app: python-kafka-producer
+    spec:
+      containers:
+        - name: python-kafka-producer
+          image: docker.io/owenerz/python-kafka-producer:latest
```
kyguy, line 19: Same as above

kyguy (commented):
Looking good @OwenCorrigan76, just have some minor nits.

As far a system testing goes, if we are going to add system tests for this and for the other client examples in general, I think that should happen in a separate PR. The yamls here are enough to prove this works IMO for now until we figure out what the system tests are going to look like

python/README.md
OwenCorrigan76: Removed

python/consumer/Dockerfile
OwenCorrigan76: No - Removed

python/consumer/Dockerfile
OwenCorrigan76: Results in the following error. It seems to reed the home directory and the variable...
`python3: can't open file '/opt/app-root/src/${HOME}kafka_consumer.py': [Errno 2] No such file or directory`

python/consumer/kafka_consumer.py
OwenCorrigan76: Yes...There is no delay in the Consumer in the Java example. It consumes at the same rate that the Producer produces, due to the sleep(1) in the producer. It prints `Waiting...` in the gaps. Should I remove the `Waiting...` message? And should I remove sleep and add the delay_ms option like the Java examples?

python/consumer/kafka_consumer_config.py
OwenCorrigan76: Had planned on using it but for now - Removed

OwenCorrigan76: See below. I am using it, I just hard coded it instead of using the variable.

python/consumer/kafka_consumer_config.py
OwenCorrigan76: Done

python/consumer/kafka_consumer_config.py
OwenCorrigan76: `'STRIMZI'` removed. `KAFKA` being used

python/producer/Dockerfile
OwenCorrigan76: Not used - removed

python/producer/Dockerfile
OwenCorrigan76: not used - removed

python/producer/Dockerfile
OwenCorrigan76: Results in the following error. It seems to reed the home directory and the variable...
`python3: can't open file '/opt/app-root/src/${HOME}kafka_consumer.py': [Errno 2] No such file or director`

python/producer/kafka_producer.py
OwenCorrigan76: Yes it was. Changing to config.delay. I've added back in DELAY_MS envVar as a String and parsed to an int and divided by 1000 (time() was in seconds) in fromEnv() to keep the YAMS consistent with Java YAMLs. e.g.
`- name: STRIMZI_DELAY_MS`
    ` value: "1000"`

python/producer/kafka_producer_config.py
OwenCorrigan76: Kafka_now used

python/producer/kafka_producer_config.py
OwenCorrigan76: Done

python/producer/requirements.txt
OwenCorrigan76: Yes

python/producer/kafka_producer_config.py
OwenCorrigan76: Done

python/python-kafka-consumer.yaml
OwenCorrigan76: Noted

python/python-kafka-deployment.yaml
OwenCorrigan76: Noted - will change to quay registry when I get nearer final submission.

python/python-kafka-deployment.yaml
OwenCorrigan76: Noted

python/python-kafka-producer.yaml
OwenCorrigan76: Noted

OwenCorrigan76:
@kyguy Git Issues with pushing old commits. Will fix

python/python-kafka-producer.yaml
```diff
@@ -0,0 +1,28 @@
...
+          env:
+            - name: STRIMZI_TOPIC
+              value: my-topic
+            - name: STRIMZI_DELAY_MS
+              value: "1000"
+            - name: KAFKA_BOOTSTRAP_SERVERS
+              value: my-cluster-kafka-bootstrap:9092
+
+
```
kyguy: extra lines

kyguy (approved):
Looks good @OwenCorrigan76. Great work!

OwenCorrigan76:
@scholzj Myself, @kyguy  and @im-konge spoke about system tests for this. It was suggested that system tests could go in a separate PR after this was merged as it doesn't seem to be as straight forward to implement as initially thought and time is against me for now. What do you think?

scholzj:
Tip: If you resolve the comments you addressed, it would be much easier to see which of them were fixed or where there is something left to do.

python/Makefile.docker
scholzj, line 1: Why do we need a separate file here instead of resuing the existing one?

scholzj (commented):
It would be great to have examples with mTLS and SASL authentication as that would be the main thing to show in these examples I think.

python/consumer/Dockerfile
OwenCorrigan76: `CMD ["python3", "kafka_consumer.py"]` works

python/kafka/producer/kafka_producer_config.py
OwenCorrigan76: `delay` reintroduced for `time.sleep(config.delay)`

python/consumer/Dockerfile
OwenCorrigan76: README also added

python/consumer/kafka_consumer_config.py
OwenCorrigan76: `client.id` added

python/Makefile.docker
OwenCorrigan76: @scholzj Thanks for the feedback. I've resolved Kyle's reviews but left some of the @mimaison unresolved until hopefully he has another pass and ok's the changes. I'll start to address your other comments re mTLS ans SASL and hopefully resolve soon. Thanks

OwenCorrigan76: @scholzj RE Makefile....I have the Makefile in the Python directory so I could just make the Python Docker images with a single command, without having to make everything. I'll strip it out though if its deemed unnecessary.

scholzj: But the `Makefile.docker` is not executed and only included, or?

scholzj: Also, you do want to build everything in a single command in some situations.

OwenCorrigan76: @scholzj Apologies..I'm not sure I understand. I know that the Makefile.docker in the Python directory is being used when I run Makefile in the Python directory. But....I've added python / producer and python / consumer to the SUBDIRS in the Makefile in the main directory. Do you think I should now remove the Makefile and Makefile.docker from the Python directory and build everything  with the one command?

scholzj: I guess the main question is ... how does it differ from [this one](https://github.com/strimzi/client-examples/blob/main/Makefile.docker)? Why do we need to have two `Makefile.docker` which basically do the same?

OwenCorrigan76: @scholzj The Makefile.docker in the Python directory is different to the Makefile.docker in the main directory in that it does not run any scripts like this: `COPY_SCRIPTS       ?= cp ../../../scripts/run.sh ./scripts/run.sh` because there are no scripts in the Python directory.
If I try to run  Makefile in the Python directory or in the main directory using this command in my python/producer/Makefile: `include ../../Makefile.docker` , to use the Makefile.docker from the main directory, it fails with the error below. Does that make sense? Is there another way I can get around this? Thanks  

```mkdir -p scripts
cp ../../../scripts/run.sh ./scripts/run.sh
cp: cannot stat '../../../scripts/run.sh': No such file or directory
make[1]: *** [../../Makefile.docker:22: docker_build] Error 1
make[1]: Leaving directory '/home/owencorrigan/Documents/client-examples/python/consumer'
make: *** [Makefile:12: python/consumer] Error 2```

Comment (scholzj) [reply python/Makefile.docker:1]: I'm not really sure what that command does. But maybe we can figure out how to unify it because having this twice does not make much sense.

Comment (OwenCorrigan76) [reply python/Makefile.docker:1]: Makefile.docker duplication resolved

File: java/http/java-http-producer/src/main/java/HttpProducer.java [hunk 42]
```diff
@@ -50,12 +53,16 @@ public static void main(String[] args) throws InterruptedException, URISyntaxExc
...
-        HttpProducer httpProducer = new HttpProducer(config);
+        HttpProducer httpProducer = new HttpProducer(config, messagesSentLatch);
         httpProducer.run();
+
+        log.info("Waiting for sending all messages");
+        messagesSentLatch.await();
     }
 
-    public HttpProducer(HttpProducerConfig config) throws URISyntaxException {
+    public HttpProducer(HttpProducerConfig config,CountDownLatch messagesSentLatch ) throws URISyntaxException {
```
ShubhamRwt:
```suggestion
    public HttpProducer(HttpProducerConfig config, CountDownLatch messagesSentLatch) throws URISyntaxException {
```

Minor typos here which can cause checkstyle to be unhappy sometimes

ShubhamRwt (approved):
Great work @OwenCorrigan76.

scholzj:
@OwenCorrigan76 What is the state of this? Should we close this until you have time to get back to it?

OwenCorrigan76:
@scholzj Yes, let's close it for now. The step I'm blocked on is to add OAuth. I'll come back to it when I can spend some time on it, thanks.

PR #115, opened 2023-03-07
Add packages to Java-Kafka directories

OwenCorrigan76:
There are missing Packages in the java-kafka folder. To bring the Client-Examples in line with other Strimzi repos, Package has been added to the Producer, Consumer and Streams directories. 
e.g.  io.strimzi.kafka.consumer.


java/common/src/main/java/io/strimzi/common/ConfigUtil.java
```diff
@@ -2,6 +2,7 @@
  * Copyright Strimzi authors.
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
+
```
scholzj: I don§t think we have an empty line here in most Strimzi files. So maybe you can here and elsewhere follow the same.

OwenCorrigan76: @scholzj Empty line removed below copyright

scholzj (approved):

ppatierno (approved):

PR #116, opened 2023-03-09
Write javadocs for new classes and methods

OwenCorrigan76:
Create javadocs for TracingSystem class and TracingInitailizer class in the java-common module in the Client-Examples repo. Also create javadocs for jaegerInitialize method and otelInitialize method in TraciningInitializer

### DIFF
--- a/java/common/src/main/java/io/strimzi/common/TracingInitializer.java
+++ b/java/common/src/main/java/io/strimzi/common/TracingInitializer.java
@@ -14,9 +14,11 @@ import io.jaegertracing.Configuration;
  */
 public class TracingInitializer {
     /**
-     Initializes Jaeger Tracing by loading the configuration from the environment
-     and registering the tracer as a global tracer
-     @return the initialized Jaeger Tracer object
+     * Initializes Jaeger Tracing by loading the configuration from the environment
+     * and registering the tracer as a global tracer
+     *
+     * @return a Jaeger Tracer object that is initialised from the environment variables
+     * using the fromEnv method
      */
     public static Tracer jaegerInitialize() {
         Tracer tracer = Configuration.fromEnv().getTracer();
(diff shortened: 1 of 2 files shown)


java/common/src/main/java/io/strimzi/common/TracingInitializer.java
```diff
@@ -9,14 +9,26 @@
...
+/**
+ * Provides methods for configuring tracing
+ */
 public class TracingInitializer {
-
+    /**
+     Initializes Jaeger Tracing by loading the configuration from the environment
+     and registering the tracer as a global tracer
+     @return the initialized Jaeger Tracer object
+     */
```
scholzj: If you add empty line between the description and the return tag it will be much more readable. ormally you would also have the `*` all along. E.g.:

```
    /**
     * Initializes Jaeger Tracing by loading the configuration from the environment
     * and registering the tracer as a global tracer
     *
     * @param xxx     Some description
     *  
     * @return the initialized Jaeger Tracer object
     */
```

When it is only return without params, you can also do somehting like this:

```
    /**
     * @return    The whole description
     */
```

OwenCorrigan76: @scholzj Is this better? Still a little unsure of how to format these properly.

java/common/src/main/java/io/strimzi/common/TracingInitializer.java
```diff
@@ -9,14 +9,30 @@
...
+ * Provides methods for configuring tracing
+ */
 public class TracingInitializer {
-
+    /**
+     * Initializes Jaeger Tracing by loading the configuration from the environment
+     * and registering the tracer as a global tracer
+     *
+     * @return a Jaeger Tracer object that is initialised from the environment variables
+     * using the fromEnv method
```
ppatierno: maybe "using the fromEnv method" is too much detailed

ppatierno (approved):
LGTM. Just a nit.

java/common/src/main/java/io/strimzi/common/TracingInitializer.java
OwenCorrigan76: @ppatierno `using the fromEnv method` removed

scholzj (approved):

PR #120, opened 2023-03-22
Update ACLs in Client Examples

OwenCorrigan76:
The old operation field in the KafkaUser resource and its ACL configuration is now deprecated. This PR updates the KafkaUser resources in the Client Examples repo to match this (to avoid the deprecation warnings).


scholzj (approved):
Thanks.

ppatierno (approved):

PR #123, opened 2023-04-03
Remove message count http examples

OwenCorrigan76:
If MessageCount is set in the YAMLs, run the Producer / Consumer for the set amount. If it is not set, run the Producer / Consumer indefinetly.

### DIFF
--- a/java/http/deployment-tracing-opentelemetry.yaml
+++ b/java/http/deployment-tracing-opentelemetry.yaml
@@ -70,8 +70,6 @@ spec:
             value: "my-group"
           - name: STRIMZI_POLL_INTERVAL
             value: "1000"
-          - name: STRIMZI_MESSAGE_COUNT
-            value: "1000000"
           - name: STRIMZI_TRACING_SYSTEM
             value: "opentelemetry"
           - name: OTEL_SERVICE_NAME
--- a/java/http/deployment-tracing-opentelemetry.yaml
+++ b/java/http/deployment-tracing-opentelemetry.yaml
@@ -26,8 +26,6 @@ spec:
             value: my-topic
           - name: STRIMZI_SEND_INTERVAL
             value: "1000"
-          - name: STRIMZI_MESSAGE_COUNT
-            value: "1000000"
           - name: STRIMZI_TRACING_SYSTEM
             value: "opentelemetry"
           - name: OTEL_SERVICE_NAME
--- a/java/http/deployment-tracing-opentelemetry.yaml
+++ b/java/http/deployment-tracing-opentelemetry.yaml
@@ -26,6 +26,8 @@ spec:
             value: my-topic
           - name: STRIMZI_SEND_INTERVAL
             value: "1000"
+          - name: STRIMZI_MESSAGE_COUNT
+            value: "1000000"
           - name: STRIMZI_TRACING_SYSTEM
             value: "opentelemetry"
           - name: OTEL_SERVICE_NAME
@@ -70,6 +72,8 @@ spec:
             value: "my-group"
           - name: STRIMZI_POLL_INTERVAL
             value: "1000"
+          - name: STRIMZI_MESSAGE_COUNT
+            value: "1000000"
           - name: STRIMZI_TRACING_SYSTEM
             value: "opentelemetry"
           - name: OTEL_SERVICE_NAME
--- a/java/http/java-http-consumer.yaml
+++ b/java/http/java-http-consumer.yaml
@@ -30,6 +30,8 @@ spec:
             value: "my-group"
           - name: STRIMZI_POLL_INTERVAL
             value: "1000"
+          - name: STRIMZI_MESSAGE_COUNT
+            value: "1000000"
         resources:
           requests:
             memory: "256Mi"
--- a/java/http/java-http-consumer.yaml
+++ b/java/http/java-http-consumer.yaml
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
       - name: java-http-consumer
-        image: quay.io/strimzi-examples/java-http-consumer:latest
+        image: docker.io/owenerz/java-http-consumer:latest
         env:
           - name: STRIMZI_HOSTNAME
             value: my-bridge-bridge-service
--- a/java/http/java-http-consumer.yaml
+++ b/java/http/java-http-consumer.yaml
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
       - name: java-http-consumer
-        image: docker.io/owenerz/java-http-consumer:latest
+        image: quay.io/strimzi-examples/java-http-consumer:latest
         env:
           - name: STRIMZI_HOSTNAME
             value: my-bridge-bridge-service
--- a/java/http/java-http-consumer.yaml
+++ b/java/http/java-http-consumer.yaml
@@ -30,8 +30,6 @@ spec:
             value: "my-group"
           - name: STRIMZI_POLL_INTERVAL
             value: "1000"
-          - name: STRIMZI_MESSAGE_COUNT
-            value: "1000000"
         resources:
           requests:
             memory: "256Mi"
--- a/java/http/java-http-consumer/src/main/java/HttpConsumer.java
+++ b/java/http/java-http-consumer/src/main/java/HttpConsumer.java
@@ -132,7 +132,6 @@ public class HttpConsumer {
         log.info("... {} messages received", this.messageReceived);
     }
 
-
     private void scheduledPoll() {
         this.poll();
         if (this.config.getMessageCount() != null && this.messageReceived == this.config.getMessageCount()) {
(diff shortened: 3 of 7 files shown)


java/http/java-http-consumer.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
       - name: java-http-consumer
-        image: quay.io/strimzi-examples/java-http-consumer:latest
+        image: docker.io/owenerz/java-http-consumer:latest
```
scholzj: This should not be here.

scholzj (commented):
I guess I'm fine with a high limit. But the Vert.x HTTP examples have no limit. So I really wonder what the limit usecase here is. Does anyone have any? If not, we should just remove it.

ppatierno:
We have different behaviours here ...
The Vert.x HTTP producer, just sends indefinitely if no message count env var is set.
I think we should follow the same logic and not setting an higher value in the YAMLs.

OwenCorrigan76:
@scholzj Considering @ppatierno 's comment, should I fix the registry directory in the consumer YAML (as outlined above) and push again with high message count?

scholzj:
Well, you need to fix the registry in any case. But I would follow Paolo's suggestion here.

scholzj:
I'm a bit lost about where we got and where we plan to go. I thought after Paolo's comment the plan was to remove the count or make it optional. But that does not seem to be reflected in the code.

ppatierno:
Agree I was expecting the same.

OwenCorrigan76:
Ah sorry. Wasn't sure whether to still remove it from the code. Now I understand. Will fix this morning.

scholzj:
I guess you don't have to remove the config option completely, just make it optional. At least that is how I would read Paolo's suggestion and if not set have it run forever.

ppatierno:
> I guess you don't have to remove the config option completely, just make it optional. At least that is how I would read Paolo's suggestion and if not set have it run forever.

Yes, that is what I mean. Option to specify should stay. I would remove the configuration of the env var in the YAML and leaving by default to send messages indefinitely.

java/http/java-http-producer/src/main/java/HttpProducer.java
```diff
@@ -64,17 +64,27 @@ public HttpProducer(HttpProducerConfig config) throws URISyntaxException {
                 GlobalOpenTelemetry.getTracer("client-examples");
     }
 
+    private Long getDefaultMessageCount() {
+        Long messageCount = this.config.getMessageCount();
+        return messageCount == null ? 1000000L : messageCount;
+    }
```
scholzj: Same as above -> why not just run forever?

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -121,19 +121,27 @@ public void subscribe() throws URISyntaxException, IOException, InterruptedExcep
     }
 
     public void run() throws InterruptedException {
+        Long messageCount = this.config.getMessageCount();
+        if ( messageCount == null ) {
+            messageCount = (Long.MAX_VALUE / this.config.getPollInterval() - 60_000L);
+        }
```
scholzj: What not just run forever when it is null?

scholzj (commented):
TBH, I'm a bit confused where we are heading with this one based on the description and on the discussion.

java/http/java-http-producer/src/main/java/HttpProducer.java
OwenCorrigan76: Honestly, I don't know how. I'll look at it again.

OwenCorrigan76: Would a while loop be appropriate here?

OwenCorrigan76: Above now removed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Above now removed

OwenCorrigan76:
Sorry this took so long. Now, using CountDownLatch, the Producer / Consumer runs for desired amount when STRIMZI_MESSAGE_COUNT is supplied, and runs forever if not supplied.

java/http/deployment-tracing-opentelemetry.yaml
```diff
@@ -26,6 +26,8 @@ spec:
             value: my-topic
           - name: STRIMZI_SEND_INTERVAL
             value: "1000"
+          - name: STRIMZI_MESSAGE_COUNT
+            value: "1000000"
```
ppatierno: Didn't we say the default behaviour will be sending forever? So why putting this so high value if now not setting the env var means sending forever?

java/http/deployment-tracing-opentelemetry.yaml
```diff
@@ -70,6 +72,8 @@ spec:
             value: "my-group"
           - name: STRIMZI_POLL_INTERVAL
             value: "1000"
+          - name: STRIMZI_MESSAGE_COUNT
+            value: "1000000"
```
ppatierno: Ditto as above but for consuming indefinitely?

java/http/deployment-tracing-opentelemetry.yaml
OwenCorrigan76: Ah, sorry. I missed this one. Fixing now.

java/http/deployment-tracing-opentelemetry.yaml
OwenCorrigan76: Removed. I forgot about the otel YAMLs. Tested without STRIMZI_MESSAGE_COUNT and working as expected.

OwenCorrigan76:
OpenTelemetry YAMLs updated.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -123,17 +129,20 @@ public void subscribe() throws URISyntaxException, IOException, InterruptedExcep
     public void run() throws InterruptedException {
         log.info("Scheduling periodic poll every {} ms waiting for {} ...", this.config.getPollInterval(), this.config.getMessageCount());
         this.executorService.scheduleAtFixedRate(this::scheduledPoll, 0, this.config.getPollInterval(), TimeUnit.MILLISECONDS);
-        this.executorService.awaitTermination(this.config.getPollInterval() * this.config.getMessageCount() + 60_000L, TimeUnit.MILLISECONDS);
         log.info("... {} messages received", this.messageReceived);
     }
 
+
```
ppatierno: why this extra line?

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -123,17 +129,20 @@ public void subscribe() throws URISyntaxException, IOException, InterruptedExcep
...
         this.poll();
-        if (this.messageReceived == this.config.getMessageCount()) {
+        if (this.config.getMessageCount() != null && this.messageReceived == this.config.getMessageCount()) {
             this.executorService.shutdown();
+            this.messagesReceivedLatch.countDown();
+            log.info("All messages received");
         }
     }
 
+
```
ppatierno: why this extra line?

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -167,11 +176,14 @@ public void poll() {
                 span.setAttribute(SemanticAttributes.HTTP_STATUS_CODE, response.statusCode());
                 span.setStatus(response.statusCode() == 200 ? StatusCode.OK : StatusCode.ERROR);
             } finally {
+
```
ppatierno: why this extra line?

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -167,11 +176,14 @@ public void poll() {
...
                 span.end();
             }
         } catch (URISyntaxException | IOException | InterruptedException e) {
             throw new RuntimeException(e);
         }
+
+
```
ppatierno: why this extra lines?

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -189,3 +201,4 @@ public void deleteConsumer() throws IOException, InterruptedException {
         }
     }
 }
+
```
ppatierno: why this extra line?

ppatierno (changes requested):

OwenCorrigan76:
@ppatierno Apologies. Extra lines were just a sloppy oversight. Removed now.

ppatierno (approved):
LGTM

scholzj (approved):

PR #126, opened 2023-05-23
Found typo in OAuth EnvVar

OwenCorrigan76:
While looking at the OAuth deployment YAML in Java/Kafka examples, in preparation to test using OAuth 0.12.0, an  error was noticed in `deployment-oauth.yaml`. The `OAUTH_CLIENT_SECRET`  environment variable key in the producer and consumer was set to `secret` instead of `clientSecret`. This error must have been included accidentally in a previous PR. This PR fixes the error.

### DIFF
--- a/java/kafka/deployment-oauth.yaml
+++ b/java/kafka/deployment-oauth.yaml
@@ -60,7 +60,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: java-kafka-producer-oauth
-                  key: secret
+                  key: clientSecret
             - name: OAUTH_TOKEN_ENDPOINT_URI
               value: https://sso.myproject.svc:8443/auth/realms/internal/protocol/openid-connect/token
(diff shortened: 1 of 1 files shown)


ppatierno (approved):

scholzj (approved):
I'm fine with this change to be merged. But I'm not sure why is it a typo. It reference a secret with some name and some key inside the secret -> what the secret name and the key is will always depend on the user, or?

ppatierno:
@scholzj I guess it's referring to what we have in the example YAML, here https://github.com/strimzi/client-examples/blob/main/java/kafka/deployment-oauth.yaml#L27
There are Secret(s) having that key to bring that data.

OwenCorrigan76:
@scholzj  Sorry, I was away from my machine for a while. `clientSecret` is what it was historically called since way back before I started. Exactly as @ppatierno said above.

scholzj:
Ok, thanks.

PR #128, opened 2023-05-29
Update OAuth to version 0.12.0

OwenCorrigan76:
Bump OAuth from version 0.8.1 to version 0.12.0. OAuth tested with Keycloak and working. OpenTelemetry and Jaeger tested and working with the updated version 0.12.0.

### DIFF
--- a/java/kafka/deployment-oauth.yaml
+++ b/java/kafka/deployment-oauth.yaml
@@ -71,7 +71,7 @@ spec:
                   name: sso-x509-https-secret
                   key: tls.crt
             - name: KAFKA_BOOTSTRAP_SERVERS
-              value: my-cluster-kafka-bootstrap:9092
+              value: my-cluster-kafka-bootstrap:9093
             - name: KAFKA_KEY_SERIALIZER
               value: "org.apache.kafka.common.serialization.StringSerializer"
             - name: KAFKA_VALUE_SERIALIZER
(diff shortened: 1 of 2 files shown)


java/kafka/deployment-oauth.yaml
```diff
@@ -219,7 +219,7 @@ spec:
                   name: sso-x509-https-secret
                   key: tls.crt
             - name: KAFKA_BOOTSTRAP_SERVERS
-              value: my-cluster-kafka-bootstrap:9093
+              value: my-cluster-kafka-bootstrap:9092
```
scholzj: Why did you change this? Given the `KAFKA_SECURITY_PROTOCOL` and `KAFKA_SSL_TRUSTSTORE_*` values I would say that 9093 fits better with our operator examples.

java/kafka/deployment-oauth.yaml
```diff
@@ -71,7 +71,7 @@ spec:
...
```
scholzj: Same as below.

OwenCorrigan76: @scholzj I had the port set at 9093 for the reason you outlined in your comment. It was only when I saw an older commit [1] (from before I started) that had set the port to 9092 that I decided to revert to 9092. I can change to 9093 again if preferred.

[1] https://github.com/strimzi/client-examples/blob/3f7a8496e03431ab2bcf9f530178e4f9cf074577/java/kafka/deployment-oauth.yaml

scholzj: In the commit you linked to, the port 9092 was correct because it was not using TLS for connecting to Kafka. But now it is using it. So now it should be 9093.

OwenCorrigan76: Ah of course. Changing now.

scholzj (approved):

ppatierno (approved):
LGTM.

Issue #118, opened 2023-03-11
Tracing seems to be enabled be default in HTTP examples


scholzj:
I deployed the HTTP consumer example with the following YAML:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: java-http-consumer
  name: java-http-consumer
spec:
  replicas: 1
  selector:
    matchLabels:
      app: java-http-consumer
  template:
    metadata:
      labels:
        app: java-http-consumer
    spec:
      containers:
      - name: java-http-consumer
        image: quay.io/strimzi-examples/java-http-consumer:latest
        env:
          - name: STRIMZI_HOSTNAME
            value: my-bridge-bridge-service
          - name: STRIMZI_PORT
            value: "8080"
          - name: STRIMZI_TOPIC
            value: kafka-test-apps
          - name: STRIMZI_GROUPID
            value: "my-group"
          - name: STRIMZI_POLL_INTERVAL
            value: "1000"
        resources:
          requests:
            memory: "256Mi"
            cpu: "250m"
          limits:
            memory: "256Mi"
            cpu: "250m"
```

So I did not enable tracing. Yet, it seems to be enabled by default as I get this in the logs:
```
Mar 11, 2023 8:07:00 PM io.opentelemetry.sdk.internal.ThrottlingLogger doLog
SEVERE: Failed to export metrics. The request could not be executed. Full error message: Failed to connect to localhost/[0:0:0:0:0:0:0:1]:4317
```

So it seems like the tracing is permanently enabled. This should be fixed and it should be enabled only when specified in the Deployment.

ppatierno:
Closed via #130

PR #130, opened 2023-06-09
Remove Otel tracing when tracingSystem set to NONE

OwenCorrigan76:
Tracing is permanently enabled even when tracingSystem is set to NONE in HttpConsumer and HttpProducer. Conditional statement added so tracing is enabled only when specified in the Deployment.


scholzj (approved):
Thanks. LGTM assuming it works.

ppatierno:
I think this deserves more investigation. I am not sure that just not setting the underneath propagator is the way to go. I mean the traces are generated but not propagated while we should understand how to not generates traces at all.
I thought that the `OpenTelemetry.noop()` returns a tracer doing nothig. We are using it in the `HttpConsumer` creation.
Even with that "noop" tracer, do we see traces? It's worth to understand how that "noop" is supposed to work than.

ppatierno:
Can you explain exactly what's the issue you see?
I just tried running the `HttpProducer` without setting the `STRIMZI_TRACING_SYSTEM` and `OTEL_SERVICE_NAME` and I don't see any trace. It seems to me that the OTel NOOP is working as expected with no changes needed.
Even just setting `OTEL_SERVICE_NAME` doesn't initialize the tracing. The code needs `STRIMZI_TRACING_SYSTEM` to be `opentelemetry`.
I did the same with the `HttpConsumer` with same result.
I don't see any issue with the current code, unless I haven't got the issue right.
All my tries were by running clients locally together with Kafka and bridge.

scholzj:
@ppatierno I raised the issue -> at least at that time, even without tracing enabled, it raised some warnings about being unable to connect somewhere to send the traces every few minutes.

scholzj:
(See #118)

OwenCorrigan76:
@ppatierno It takes around 50 - 70 messages before any tracing logs appear but they are there, even with `Opentelemetry.noop`  in use. I've been testing it locally and in a cluster with the bridge from Minikube and this appears in the logs, even when tracing is not enabled:
`Jun 12, 2023 8:04:37 AM io.opentelemetry.sdk.internal.ThrottlingLogger doLog                       │
│ SEVERE: Failed to export metrics. The request could not be executed.`


While I don't know why `noop` does not stop the tracing completely, the conditional statement I added has the desired effect of removing the openTelemetry log when tracing is not set.

ppatierno:
I tried with a longer run and it happened after around 50 messages. I guess there should be a 60 seconds timeout for OpenTelemetry (considering I was sending one message per second).
While your solution could fix the problem, I am not still sure it's the right way to go. Did you engage with the OpenTelemetry community to ask the purpose of the Noop tracer and why it doesn't work somehow?

OwenCorrigan76:
@ppatierno I haven't, as I thought you might know. I will engage now and see what I can find out.

ppatierno:
I read the error better now   ... `Failed to export metrics` it's not about tracing!
Metrics and tracing in OpenTelemetry are two different things.
Metrics export is enabled by default in OTel, you need to do what we have in the bridge.

```java
System.setProperty("otel.metrics.exporter", "none"); // disable metrics
```

You can set use this line in the `HttpConsumer` and `HttpProducer` constructors.

java/http/java-http-producer/src/main/java/HttpProducer.java
```diff
@@ -44,6 +44,7 @@ public static void main(String[] args) throws InterruptedException, URISyntaxExc
 
         HttpProducerConfig config = HttpProducerConfig.fromEnv();
         CountDownLatch messagesSentLatch = new CountDownLatch(1);
+
```
ppatierno: why this additional line?

java/http/java-http-producer/src/main/java/HttpProducer.java
```diff
@@ -122,4 +124,3 @@ private void send() {
         }
     }
 }
-
```
ppatierno, line 125: why removing this line?

OwenCorrigan76: Put back one empty line. Should be correct now.

java/http/java-http-producer/src/main/java/HttpProducer.java
OwenCorrigan76: Accidental. Now removed.

ppatierno (approved):
LGTM. Assuming you tried on your side as well. It worked for me.

scholzj (approved):
So, should this be added to the other examples as well?

scholzj (commented):
So, should this be added to the other examples as well?

OwenCorrigan76:
@ppatierno Yes, just tested again and working as expected (no metrics exported logs). @scholzj I never noticed this `failed to export metrics` log in any of the other examples. Do you mean the Vert.x example?

scholzj:
> I never noticed this failed to export metrics log in any of the other examples. Do you mean the Vert.x example?

All the other examples use OpenTelemetry as well. So why is it not needed there as well?

OwenCorrigan76:
@scholzj I'll deploy the Vertx and Java/Kafka Producer / Consumer and investigate whether this is an issue.

ppatierno:
@OwenCorrigan76 @scholzj AFAICS the Vert.x examples are still using OpenTracing. No OpenTelemetry at all there, so we should not have any issue. The only issue is that we should migrate them to use OpenTelemetry soon when all Strimzi components will get rid of OpenTelemetry.
Regarding the Kafka examples, they use the OTel Kafka instrumentation library, not sure if it does help us because the OTel SDK is not directly involved. Of course, it would need testing to confirm or not.

ppatierno:
@OwenCorrigan76 @scholzj AFAICS the Vert.x examples are still using OpenTracing. No OpenTelemetry at all there, so we should not have any issue. The only issue is that we should migrate them to use OpenTelemetry soon when all Strimzi components will get rid of OpenTelemetry.
Regarding the Kafka examples, they use the OTel Kafka instrumentation library, not sure if it does help us because the OTel SDK is not directly involved. Of course, it would need testing to confirm or not.

ppatierno:
@OwenCorrigan76 @scholzj AFAICS the Vert.x examples are still using OpenTracing. No OpenTelemetry at all there, so we should not have any issue. The only issue is that we should migrate them to use OpenTelemetry soon when all Strimzi components will get rid of OpenTelemetry.
Regarding the Kafka examples, they use the OTel Kafka instrumentation library, not sure if it does help us because the OTel SDK is not directly involved. Of course, it would need testing to confirm or not.

OwenCorrigan76:
@ppatierno Sorry, I forgot that Vert.x is only using OpenTracing until I went back to look at the example:
``` import io.opentracing.Span;
import io.opentracing.SpanContext;
import io.opentracing.Tracer;
import io.opentracing.propagation.Format;
import io.opentracing.propagation.TextMapAdapter;
import io.opentracing.tag.Tags;
import io.opentracing.util.GlobalTracer; 
```


I've been running the Java/ Kafka examples now for around 5 minutes with NO additional logs around metrics.

scholzj (approved):

PR #131, opened 2023-06-15
Add Spotbugs to Maven and Fix Spotbugs Warnings

OwenCorrigan76:
Spotbugs added to Client Examples. As a result of Spotbug warnings, I have edited the files with warnings as best as i could to get rid of the warnings but ensure that the code still works properly. Each component now compiles and runs as expected and has no Spotbugs warnings. I suspect there will be some comments on how my changes may have been carried out more efficiently.


scholzj (commented):
This looks good. But you should also add it to the CI in https://github.com/strimzi/client-examples/tree/main/.azure -> so that it is actually executed in the CI pipeline.

OwenCorrigan76:
@scholzj I've added the following to the `build-pipeline.yaml` under jobs/ steps/ 
```
- script: 'make spotbugs'
  displayName: 'Run SpotBugs' 
```
Is this sufficient or do i need to add anything else in the Azure directory? Thanks

scholzj:
@OwenCorrigan76 I think that might be fine -> push it in and we will see.

pom.xml
```diff
@@ -26,9 +26,12 @@
         <maven.shade.version>3.4.1</maven.shade.version>
         <maven.checkstyle.version>3.2.0</maven.checkstyle.version>
 
+
+
```
scholzj: Not needed I guess?

python/kafka/producer/__pycache__/kafka_producer_config.cpython-310.pyc
scholzj, line 1: This file probably should not be there?

scholzj (approved):
Two more nits to fix. But LGTM otherwise.

pom.xml
OwenCorrigan76: Extra lines removed

python/kafka/producer/__pycache__/kafka_producer_config.cpython-310.pyc
OwenCorrigan76: Removed

scholzj (approved):

java/kafka/pom.xml
```diff
@@ -18,4 +18,16 @@
...
+    <plugin>
+        <groupId>com.github.spotbugs</groupId>
+        <artifactId>spotbugs-maven-plugin</artifactId>
+        <configuration>
+            <!-- Configures the file for excluding warnings -->
+            <excludeFilterFile>${project.basedir}/../../../.spotbugs/spotbugs-exclude.xml</excludeFilterFile>
+        </configuration>
+    </plugin>
+    </plugins>
+    </build>
```
ppatierno, line 32: wrong indentation of plugins and plugin tags

ppatierno (approved):
Just one nit but LGTM. No need to re-review when it's ok. We can just merge.

java/kafka/pom.xml
OwenCorrigan76: @ppatierno Indentation fixed

Issue #119, opened 2023-03-11
HTTP Example consumer should commit messages


scholzj:
The HTTP Example consumer does not seem to commit the messages it consumes. I think adding that would be a good part of the example.

ppatierno:
Closed via #132

PR #132, opened 2023-06-26
Add commits to Java Http Consumer

OwenCorrigan76:
Committing consumer messages and commit message logs as part of the Java Http Consumer client. This is outlined in the following issue [1].

[1] https://github.com/strimzi/client-examples/issues/119

### DIFF
--- a/.checkstyle/checkstyle.xml
+++ b/.checkstyle/checkstyle.xml
@@ -95,7 +95,9 @@
         <module name="ParameterNumber"/>
         <module name="ClassDataAbstractionCoupling"/>
         <module name="BooleanExpressionComplexity"/>
-        <module name="ClassFanOutComplexity"/>
+        <module name="ClassFanOutComplexity">
+        <property name="max" value="22"/>
+        </module>
         <module name="CyclomaticComplexity"/>
         <module name="JavaNCSS"/>
         <module name="NPathComplexity"/>
--- a/.checkstyle/checkstyle.xml
+++ b/.checkstyle/checkstyle.xml
@@ -95,9 +95,7 @@
         <module name="ParameterNumber"/>
         <module name="ClassDataAbstractionCoupling"/>
         <module name="BooleanExpressionComplexity"/>
-        <module name="ClassFanOutComplexity">
-        <property name="max" value="22"/>
-        </module>
+        <module name="ClassFanOutComplexity"/>
         <module name="CyclomaticComplexity"/>
         <module name="JavaNCSS"/>
         <module name="NPathComplexity"/>
--- a/.spotbugs/spotbugs-exclude.xml
+++ b/.spotbugs/spotbugs-exclude.xml
@@ -6,7 +6,4 @@
     <Match>
         <Bug pattern="EI_EXPOSE_REP2"/>
     </Match>
-    <Match>
-        <Bug pattern="DM_CONVERT_CASE"/>
-    </Match>
 </FindBugsFilter>
--- a/.spotbugs/spotbugs-exclude.xml
+++ b/.spotbugs/spotbugs-exclude.xml
@@ -6,4 +6,10 @@
     <Match>
         <Bug pattern="EI_EXPOSE_REP2"/>
     </Match>
+    <Match>
+        <Bug pattern="DM_CONVERT_CASE"/>
+    </Match>
+    <Match>
+        <Bug pattern="ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD"/>
+    </Match>
 </FindBugsFilter>
--- a/.spotbugs/spotbugs-exclude.xml
+++ b/.spotbugs/spotbugs-exclude.xml
@@ -9,7 +9,4 @@
     <Match>
         <Bug pattern="DM_CONVERT_CASE"/>
     </Match>
-    <Match>
-        <Bug pattern="ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD"/>
-    </Match>
 </FindBugsFilter>
--- a/java/http/java-http-consumer/pom.xml
+++ b/java/http/java-http-consumer/pom.xml
@@ -52,11 +52,6 @@
             <groupId>io.opentelemetry</groupId>
             <artifactId>opentelemetry-semconv</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka-clients</artifactId>
-            <version>3.4.1</version>
-        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/java/http/java-http-consumer/pom.xml
+++ b/java/http/java-http-consumer/pom.xml
@@ -52,8 +52,12 @@
             <groupId>io.opentelemetry</groupId>
             <artifactId>opentelemetry-semconv</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
+            <version>3.4.1</version>
+        </dependency>
     </dependencies>
-
     <build>
         <plugins>
             <plugin>
--- a/java/http/java-http-consumer/src/main/java/HttpConsumer.java
+++ b/java/http/java-http-consumer/src/main/java/HttpConsumer.java
@@ -52,7 +52,7 @@ public class HttpConsumer {
     private ScheduledExecutorService executorService;
     private Tracer tracer;
     private CountDownLatch messagesReceivedLatch;
-    private static boolean enableAutoCommit = false;
+    private boolean enableAutoCommit = false;
 
     private static List<String> commonProps;
     static {
@@ -68,7 +68,6 @@ public class HttpConsumer {
     public static void main(String[] args) throws URISyntaxException, IOException, InterruptedException {
         HttpConsumerConfig config = HttpConsumerConfig.fromEnv();
         CountDownLatch messagesReceivedLatch = new CountDownLatch(1);
-        Properties props = config.getProperties();
 
         TracingSystem tracingSystem = config.getTracingSystem();
         if (tracingSystem != TracingSystem.NONE) {
@@ -80,7 +79,6 @@ public class HttpConsumer {
         }
 
         HttpConsumer consumer = new HttpConsumer(config, messagesReceivedLatch);
-        enableAutoCommit = (props.getProperty(ENABLE_AUTO_COMMIT_CONFIG)) == null || Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
 
         try {
             consumer.createConsumer();
@@ -104,7 +102,7 @@ public class HttpConsumer {
                 OpenTelemetry.noop().getTracer("client-examples") :
                 GlobalOpenTelemetry.getTracer("client-examples");
     }
-    @SuppressWarnings("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD")
+
     public void createConsumer() throws IOException, InterruptedException, URISyntaxException {
         Properties props = config.getProperties();
         enableAutoCommit = (props.getProperty(ENABLE_AUTO_COMMIT_CONFIG)) == null || Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
@@ -192,7 +190,6 @@ public class HttpConsumer {
         log.info("Scheduling periodic poll every {} ms waiting for {} ...", this.config.getPollInterval(), this.config.getMessageCount());
         this.executorService.scheduleAtFixedRate(this::scheduledPoll, 0, this.config.getPollInterval(), TimeUnit.MILLISECONDS);
         log.info("... {} messages received", this.messageReceived);
-
     }
 
     private void scheduledPoll() {
--- a/java/http/java-http-consumer/src/main/java/HttpConsumer.java
+++ b/java/http/java-http-consumer/src/main/java/HttpConsumer.java
@@ -31,6 +31,7 @@ import java.util.HashMap;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Locale;
 import java.util.Properties;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executors;
@@ -105,7 +106,7 @@ public class HttpConsumer {
 
     public void createConsumer() throws IOException, InterruptedException, URISyntaxException {
         Properties props = config.getProperties();
-        enableAutoCommit = (props.getProperty(ENABLE_AUTO_COMMIT_CONFIG)) == null || Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
+        enableAutoCommit = Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
         ObjectMapper objectMapper = new ObjectMapper();
         Map<String, Object> map = new HashMap<>();
         map.put("name", "my-consumer");
@@ -171,9 +172,8 @@ public class HttpConsumer {
         }
     }
 
-    @SuppressWarnings({"Regexp"})
     public static boolean isStringBoolean(String input) {
-        String lowerCaseInput = input.toLowerCase();
+        String lowerCaseInput = input.toLowerCase(Locale.ENGLISH);
         return lowerCaseInput.equals("true") || lowerCaseInput.equals("false");
     }
 
--- a/java/http/java-http-consumer/src/main/java/HttpConsumer.java
+++ b/java/http/java-http-consumer/src/main/java/HttpConsumer.java
@@ -27,10 +27,13 @@ import java.net.URISyntaxException;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
+import java.util.Properties;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+
+import static org.apache.kafka.clients.consumer.ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG;
 
 public class HttpConsumer {
 
@@ -45,11 +48,13 @@ public class HttpConsumer {
     private ScheduledExecutorService executorService;
     private Tracer tracer;
     private CountDownLatch messagesReceivedLatch;
+    private boolean manualCommit = false;
 
 
     public static void main(String[] args) throws URISyntaxException, IOException, InterruptedException {
         HttpConsumerConfig config = HttpConsumerConfig.fromEnv();
         CountDownLatch messagesReceivedLatch = new CountDownLatch(1);
+        Properties props = config.getProperties();
 
         TracingSystem tracingSystem = config.getTracingSystem();
         if (tracingSystem != TracingSystem.NONE) {
@@ -61,6 +66,8 @@ public class HttpConsumer {
         }
 
         HttpConsumer consumer = new HttpConsumer(config, messagesReceivedLatch);
+        boolean enableAutoCommit = (props.getProperty(ENABLE_AUTO_COMMIT_CONFIG)) == null || Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
+        consumer.manualCommit = !enableAutoCommit;
         try {
             consumer.createConsumer();
             consumer.subscribe();
@@ -167,14 +174,14 @@ public class HttpConsumer {
 
                 ArrayNode records = (ArrayNode) MAPPER.readTree(response.body());
                 log.info("... got {} records", records.size());
-                if (records.size() > 0) {
-                    this.commitMessages();
-                }
+
                 for (JsonNode record : records) {
                     log.info("Record {}", record);
                     this.messageReceived++;
                 }
-
+                if (records.size() > 0 && this.manualCommit) {
+                    this.commitMessages();
+                }
                 span.setAttribute(SemanticAttributes.HTTP_STATUS_CODE, response.statusCode());
                 span.setStatus(response.statusCode() == 200 ? StatusCode.OK : StatusCode.ERROR);
             } finally {
--- a/java/http/java-http-consumer/src/main/java/HttpConsumer.java
+++ b/java/http/java-http-consumer/src/main/java/HttpConsumer.java
@@ -27,18 +27,22 @@ import java.net.URISyntaxException;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
+import java.util.HashMap;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
-import static org.apache.kafka.clients.consumer.ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG;
-
+@SuppressWarnings("classFanOutComplexity")
 public class HttpConsumer {
 
     private static final Logger log = LogManager.getLogger(HttpConsumer.class);
     private static final ObjectMapper MAPPER = new ObjectMapper();
+    private static final String ENABLE_AUTO_COMMIT_CONFIG = "enable.auto.commit";
 
     private final HttpConsumerConfig config;
     private HttpClient httpClient;
@@ -48,8 +52,18 @@ public class HttpConsumer {
     private ScheduledExecutorService executorService;
     private Tracer tracer;
     private CountDownLatch messagesReceivedLatch;
-    private boolean manualCommit = false;
+    private static boolean enableAutoCommit = false;
 
+    private static List<String> commonProps;
+    static {
+        commonProps = new ArrayList<>();
+        commonProps.add("auto.offset.reset");
+        commonProps.add("enable.auto.commit");
+        commonProps.add("fetch.min.bytes");
+        commonProps.add("request.timeout.ms");
+        commonProps.add("client.id");
+        commonProps.add("isolation.level");
+    }
 
     public static void main(String[] args) throws URISyntaxException, IOException, InterruptedException {
         HttpConsumerConfig config = HttpConsumerConfig.fromEnv();
@@ -66,8 +80,8 @@ public class HttpConsumer {
         }
 
         HttpConsumer consumer = new HttpConsumer(config, messagesReceivedLatch);
-        boolean enableAutoCommit = (props.getProperty(ENABLE_AUTO_COMMIT_CONFIG)) == null || Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
-        consumer.manualCommit = !enableAutoCommit;
+        enableAutoCommit = (props.getProperty(ENABLE_AUTO_COMMIT_CONFIG)) == null || Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
+
         try {
             consumer.createConsumer();
             consumer.subscribe();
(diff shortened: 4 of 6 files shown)


java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -166,7 +167,9 @@ public void poll() {
 
                 ArrayNode records = (ArrayNode) MAPPER.readTree(response.body());
                 log.info("... got {} records", records.size());
-
+                if (records.size() > 0) {
+                    this.commitMessages();
+                }
```
ppatierno: In general you should commit after the messages processing, where in our case processing means just printing in the log.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -131,6 +131,7 @@ public void run() throws InterruptedException {
         log.info("Scheduling periodic poll every {} ms waiting for {} ...", this.config.getPollInterval(), this.config.getMessageCount());
         this.executorService.scheduleAtFixedRate(this::scheduledPoll, 0, this.config.getPollInterval(), TimeUnit.MILLISECONDS);
         log.info("... {} messages received", this.messageReceived);
+
```
ppatierno: remove blank line

ppatierno (commented):
I left two comments but I have a wider one ...
In our example, the HTTP consumer is not created by specifying the `enable.auto.commit: false` so it means it's using the default one ... which is `true`. The underlying Kafka consumer is committing automatically every `auto.commit.interval.ms` (5 secs by default).
If we do really want to have manual commit, we should create the HTTP consumer with auto commit disabled or we should make it configurable (which is something we have in the `KafkaConsumerExample`).

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: @ppatierno Does this mean just moving the `commitMessages()` call below the for loop [1] in the `poll` method? 
```  
for (JsonNode record : records) {
                    log.info("Record {}", record);
                    this.messageReceived++;
                } 
```

ppatierno: related to the specific comment yes, but if you read the main comment I left on the review, it's useless.

OwenCorrigan76: @ppatierno Examining the Java Kafka example `enable auto commit` to implement here

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: removed

OwenCorrigan76:
@ppatierno When no environment variable set in the yaml, there is no commit logs on the bridge or the consumer. With `KAFKA_ENABLE_AUTO_OFFESET` set to `false` in the yaml, there are commit messages and logs in the consumer and the bridge. Do I have this the right way around? Thanks

.checkstyle/checkstyle.xml
```diff
@@ -95,7 +95,9 @@
         <module name="ParameterNumber"/>
         <module name="ClassDataAbstractionCoupling"/>
         <module name="BooleanExpressionComplexity"/>
-        <module name="ClassFanOutComplexity"/>
+        <module name="ClassFanOutComplexity">
+        <property name="max" value="22"/>
+        </module>
```
ppatierno: I would suppress the warning where it happens instead of setting the "right" max value here.

java/http/java-http-consumer/pom.xml
```diff
@@ -52,8 +52,12 @@
             <groupId>io.opentelemetry</groupId>
             <artifactId>opentelemetry-semconv</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
+            <version>3.4.1</version>
+        </dependency>
```
ppatierno: I see you are adding this Kafka dependency just for getting the constant for the `enable.auto.commit` configuration property. Tbh I would avoid that, just for a constant. Declare your own in the code and remove the dependency I would say.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -61,6 +66,8 @@ public static void main(String[] args) throws URISyntaxException, IOException, I
         }
 
         HttpConsumer consumer = new HttpConsumer(config, messagesReceivedLatch);
+        boolean enableAutoCommit = (props.getProperty(ENABLE_AUTO_COMMIT_CONFIG)) == null || Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
+        consumer.manualCommit = !enableAutoCommit;
```
ppatierno: why do you need two variables with one the opposite of the other? You can just define the `enableAutoCommit` as the member field (instead of the `manualCommit`) but avoiding to do this negation.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -171,7 +179,9 @@ public void poll() {
                     log.info("Record {}", record);
                     this.messageReceived++;
                 }
-
+                if (records.size() > 0 && this.manualCommit) {
+                    this.commitMessages();
+                }
```
ppatierno, line 225: you are committing messages if manualCommit (or !enableAutoCommit, see my previous comment), but still the underlying Kafka consumer on the bridge is configured by auto committing because it's the default for Kafka (see https://kafka.apache.org/documentation/#consumerconfigs_enable.auto.commit).
I don't want to spoiler everything but remember my comment ...

> If we do really want to have manual commit, we should create the HTTP consumer with auto commit disabled or we should make it configurable (which is something we have in the KafkaConsumerExample).

What I see you are adding the `KAFKA_` related stuff from the Kafka examples but they are not used at all to configure the corresponding Kafka consumer via the HTTP consumer creation. You are using the properties just for getting the auto commit configuration.
Using those properties is simple for a Kafka based example because the creation of the Kafka consumer just gets ... a `Properties` instance, which is not true for an HTTP consumer creation.
See https://strimzi.io/docs/bridge/latest/#proc-creating-kafka-bridge-consumer-bridge

With your current code, what happens if the user specifies a KAFKA_AUTO_OFFSET_RESET for example? Are you just ignoring it? Should it configure the corresponding HTTP consumer?

ppatierno (changes requested):

.checkstyle/checkstyle.xml
OwenCorrigan76: Done

java/http/java-http-consumer/pom.xml
OwenCorrigan76: Done

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: This seems to be resolved now by creating a new JSON with the configs payload. Hashmap is used though ObjectNode may be a better route.

.spotbugs/spotbugs-exclude.xml
```diff
@@ -6,4 +6,10 @@
     <Match>
         <Bug pattern="EI_EXPOSE_REP2"/>
     </Match>
+    <Match>
+        <Bug pattern="DM_CONVERT_CASE"/>
+    </Match>
+    <Match>
+        <Bug pattern="ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD"/>
+    </Match>
```
ppatierno: I guess you had to add this because of the usage of enableAutoCommit in the main (which is static) and an instance method. I would remove this and follow the comment I have later in this review.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -83,12 +104,21 @@ public HttpConsumer(HttpConsumerConfig config, CountDownLatch messagesReceivedLa
                 OpenTelemetry.noop().getTracer("client-examples") :
                 GlobalOpenTelemetry.getTracer("client-examples");
     }
-
+    @SuppressWarnings("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD")
```
ppatierno: As well you don't need this one (in general you don't need both suppress warnings and an exclusion in the XML file)

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -45,11 +52,23 @@ public class HttpConsumer {
     private ScheduledExecutorService executorService;
     private Tracer tracer;
     private CountDownLatch messagesReceivedLatch;
+    private static boolean enableAutoCommit = false;
```
ppatierno: You don't need this field to be static. I see you are actually reading it in the main, so the reason to make it static but it's not needed.
In the end, it has just to be a non static field of the HttpConsumer class which gets it from the config (and related properties collection). This field is used in all instance methods of this class, there is no usage in the main so no need to write it there.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -61,6 +80,8 @@ public static void main(String[] args) throws URISyntaxException, IOException, I
         }
 
         HttpConsumer consumer = new HttpConsumer(config, messagesReceivedLatch);
+        enableAutoCommit = (props.getProperty(ENABLE_AUTO_COMMIT_CONFIG)) == null || Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
+
```
ppatierno: What's the purpose of setting the enableAutoCommit here? You are already good by reading it in the consumer creation method so you can make this field non static and remove the suppress warnings. See my other comments.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -127,10 +173,26 @@ public void subscribe() throws URISyntaxException, IOException, InterruptedExcep
         }
     }
 
+    @SuppressWarnings({"Regexp"})
```
ppatierno: Which error are you getting for needing this?

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -127,10 +173,26 @@ public void subscribe() throws URISyntaxException, IOException, InterruptedExcep
...
+        } catch (NumberFormatException e) {
+            return false;
+        }
+    }
+
     public void run() throws InterruptedException {
         log.info("Scheduling periodic poll every {} ms waiting for {} ...", this.config.getPollInterval(), this.config.getMessageCount());
         this.executorService.scheduleAtFixedRate(this::scheduledPoll, 0, this.config.getPollInterval(), TimeUnit.MILLISECONDS);
         log.info("... {} messages received", this.messageReceived);
+
```
ppatierno: No need for this blank line.

java/kafka/consumer/pom.xml
```diff
@@ -52,6 +48,10 @@
             <groupId>io.strimzi</groupId>
             <artifactId>kafka-oauth-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
+        </dependency>
```
ppatierno: Why moving this dependency in the Kafka related examples? I don't think it's needed.

ppatierno (changes requested):

.spotbugs/spotbugs-exclude.xml
OwenCorrigan76: Done

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Removed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: `Line matches the illegal pattern '\.to(Lower|Upper)Case\(\)'. [Regexp]`

java/kafka/consumer/pom.xml
OwenCorrigan76:
``` import org.apache.kafka.clients.consumer.ConsumerConfig;
import org.apache.kafka.clients.consumer.ConsumerRecord;
import org.apache.kafka.clients.consumer.ConsumerRecords;
import org.apache.kafka.clients.consumer.KafkaConsumer; 
```

It is being used extensively in KafkaConsumeExample

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Removed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Fixed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Removed

java/kafka/consumer/pom.xml
ppatierno: I know it's used, it's the Kafka based example but AFAICS you moved from one place to another and in a PR which is not related to the Kafka examples.

ppatierno: You moved the same block from line 19 to line 51 for now reasons :-) This is the point I am making.

.spotbugs/spotbugs-exclude.xml
```diff
@@ -6,4 +6,7 @@
     <Match>
         <Bug pattern="EI_EXPOSE_REP2"/>
     </Match>
+    <Match>
+        <Bug pattern="DM_CONVERT_CASE"/>
+    </Match>
```
ppatierno: I think we should remove this as well by using `toLowerCase(Locale.ENGLISH)` where you are lower casing a string.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -127,6 +171,21 @@ public void subscribe() throws URISyntaxException, IOException, InterruptedExcep
         }
     }
 
+    @SuppressWarnings({"Regexp"})
+    public static boolean isStringBoolean(String input) {
+        String lowerCaseInput = input.toLowerCase();
+        return lowerCaseInput.equals("true") || lowerCaseInput.equals("false");
+    }
```
ppatierno, line 168: I think we can remove this method and using `Boolean.parseBoolean(<string>)`. Its internal implementation return true if the string is "true", otherwise return false, which is good enough.

.spotbugs/spotbugs-exclude.xml
ppatierno: Actually I made another suggestion to get rid of the entire method. See later.

ppatierno (changes requested):

java/http/java-http-consumer/src/main/java/HttpConsumer.java
ppatierno: Maybe following the next comment we can get rid of the entire method.

ppatierno (changes requested):

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -85,10 +104,19 @@ public HttpConsumer(HttpConsumerConfig config, CountDownLatch messagesReceivedLa
     }
 
     public void createConsumer() throws IOException, InterruptedException, URISyntaxException {
-        String consumerInfo = "{\"name\":\"" + this.config.getClientId() + "\",\"format\":\"json\",\"auto.offset.reset\":\"earliest\"}";
+        Properties props = config.getProperties();
+        enableAutoCommit = (props.getProperty(ENABLE_AUTO_COMMIT_CONFIG)) == null || Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
```
ppatierno: I don't think we need a check on `null` the `parseBoolean` would take care for you about it.

java/kafka/consumer/pom.xml
OwenCorrigan76: Ah sorry. Now I get you. I must have deleted it from the wrong pom and then realised I had to import it again and it was placed in a different spot. That's why it showed up in the PR. Putting back to the original place.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
ppatierno: Thinking more, you are using this method to understand if the string is a boolean or not but you are not interested in the value itself, so you need `true` if the string is "true" or "false", not when it's just "true" as it would happen for `parseBoolean`. So take this method but fix my other comments on the lower case and warning.

OwenCorrigan76: Done

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Removed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -61,6 +80,7 @@ public static void main(String[] args) throws URISyntaxException, IOException, I
         }
 
         HttpConsumer consumer = new HttpConsumer(config, messagesReceivedLatch);
+
```
kyguy: extra line

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -108,6 +137,22 @@ public void createConsumer() throws IOException, InterruptedException, URISyntax
...
+        if (props.stringPropertyNames().contains(k)) {
+            String v = props.getProperty(k);
+            Object obj;
+            if (isStringBoolean(v)) {
+                obj = Boolean.parseBoolean(v);
+            } else if (isStringNumber(v)) {
+                obj = Integer.parseInt(v);
+            } else {
+                obj = v;
+            }
```
ppatierno: I would avoid the else branch, and just doing `Object obj = v;` at the beginning.
It will be overwrite if needed or stay as `v` in the end.

ppatierno (commented):
We are almost there ... :-)

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -85,10 +105,19 @@ public HttpConsumer(HttpConsumerConfig config, CountDownLatch messagesReceivedLa
     }
 
     public void createConsumer() throws IOException, InterruptedException, URISyntaxException {
-        String consumerInfo = "{\"name\":\"" + this.config.getClientId() + "\",\"format\":\"json\",\"auto.offset.reset\":\"earliest\"}";
+        Properties props = config.getProperties();
+        enableAutoCommit = Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
+        ObjectMapper objectMapper = new ObjectMapper();
```
kyguy: Personally, would move this to the line above consumerInfo

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -108,6 +137,22 @@ public void createConsumer() throws IOException, InterruptedException, URISyntax
...
+            Object obj;
+            if (isStringBoolean(v)) {
+                obj = Boolean.parseBoolean(v);
+            } else if (isStringNumber(v)) {
+                obj = Integer.parseInt(v);
+            } else {
+                obj = v;
+            }
+            map.put(k, obj);
+            log.info("key: {}, value: {}", k, v);
```
kyguy: I would remove this line since they will already be printed out with consumerInfo print line and it doesn't print the "name" and "format" configurations

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -85,10 +105,19 @@ public HttpConsumer(HttpConsumerConfig config, CountDownLatch messagesReceivedLa
     }
 
     public void createConsumer() throws IOException, InterruptedException, URISyntaxException {
-        String consumerInfo = "{\"name\":\"" + this.config.getClientId() + "\",\"format\":\"json\",\"auto.offset.reset\":\"earliest\"}";
+        Properties props = config.getProperties();
+        enableAutoCommit = Boolean.parseBoolean(props.getProperty(ENABLE_AUTO_COMMIT_CONFIG));
```
kyguy: I would set this in the HttpConsumerConfig after properties.putAll line in the fromEnv() method, adding enableAuthCommit as a field to HttpConsumerConfig class

kyguy (approved):
Looks great @OwenCorrigan76, just a few minor comments

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Refactored

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: removed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Makes sense. Done

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: removed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Done

java/http/java-http-consumer/src/main/java/HttpConsumer.java
kyguy: extra line is still here

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -85,10 +103,19 @@ public HttpConsumer(HttpConsumerConfig config, CountDownLatch messagesReceivedLa
     }
 
     public void createConsumer() throws IOException, InterruptedException, URISyntaxException {
-        String consumerInfo = "{\"name\":\"" + this.config.getClientId() + "\",\"format\":\"json\",\"auto.offset.reset\":\"earliest\"}";
+        Properties props = config.getProperties();
+        enableAutoCommit = config.getEnableAutoCommit();
+        Map<String, Object> map = new HashMap<>();
```
kyguy:
```suggestion
        Map<String, Object> consumerConfig = createConsumerConfig(props);
```

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -85,10 +103,19 @@ public HttpConsumer(HttpConsumerConfig config, CountDownLatch messagesReceivedLa
...
+        map.put("name", "my-consumer");
+        map.put("format", "json");
+        for (String k : commonProps) {
+            processProperty(props, k, map);
+        }
```
kyguy, line 118: Move the above 5 lines to createConsumerConfig() method

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -108,6 +135,19 @@ public void createConsumer() throws IOException, InterruptedException, URISyntax
...
+            Object obj = v;
+            if (isStringBoolean(v)) {
+                obj = Boolean.parseBoolean(v);
+            } else if (isStringNumber(v)) {
+                obj = Integer.parseInt(v);
+            }
+            map.put(k, obj);
+        }
+    }
+
```
kyguy: I would change the `processProperty` method into `createConsumerConfig` method like this
```suggestion
  public void createConsumerConfig(Properties props) {
    Map<String, Object> consumerConfig = new HashMap<>();
    consumerConfig.put("name", "my-consumer");
    consumerConfig.put("format", "json");
    for (String k : commonProps) {
        if (props.stringPropertyNames().contains(k)) {
            String v = props.getProperty(k);
            Object obj = v;
            if (isStringBoolean(v)) {
                obj = Boolean.parseBoolean(v);
            } else if (isStringNumber(v)) {
                obj = Integer.parseInt(v);
            }
            consumerConfig.put(k, obj);
        }
      }
      return consumerConfig;
  }
```

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -85,10 +103,19 @@ public HttpConsumer(HttpConsumerConfig config, CountDownLatch messagesReceivedLa
     }
 
     public void createConsumer() throws IOException, InterruptedException, URISyntaxException {
-        String consumerInfo = "{\"name\":\"" + this.config.getClientId() + "\",\"format\":\"json\",\"auto.offset.reset\":\"earliest\"}";
+        Properties props = config.getProperties();
+        enableAutoCommit = config.getEnableAutoCommit();
```
kyguy: This line can be removed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -171,7 +225,9 @@ public void poll() {
                     log.info("Record {}", record);
                     this.messageReceived++;
                 }
-
+                if (records.size() > 0 && !enableAutoCommit) {
```
kyguy:
```suggestion
                if (records.size() > 0 && !config.getEnableAutoCommit();) {
```

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -45,7 +51,18 @@ public class HttpConsumer {
     private ScheduledExecutorService executorService;
     private Tracer tracer;
     private CountDownLatch messagesReceivedLatch;
+    private boolean enableAutoCommit = false;
```
kyguy: This line can be removed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
ppatierno: It's not just about the `processProperty` as it is today but I see that the proposal from @kyguy includes the loop on the `commonProps` as well. I like the idea.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
ppatierno: Good catch Kyle, then `enableAutoCommit` can be removed.

ppatierno (commented):
@OwenCorrigan76 I see some pending comments from Kyle which need to be fixed Then I will have another (maybe final) pass.

OwenCorrigan76:
@ppatierno Thanks Paolo. Looking at Kyle's comments now.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Extra line removed. Now same as original.
`createConsumerConfig` created

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Now removed. Same as original

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Moved into `createConsumerConfig()`

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Refactored. `  public Map<String, Object> createConsumerConfig(Properties props) ` Map return type, not void.

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Removed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Removed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
OwenCorrigan76: Removed

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -84,11 +100,31 @@ public HttpConsumer(HttpConsumerConfig config, CountDownLatch messagesReceivedLa
...
+            }
+        }
+        return consumerConfig;
+    }
+
     public void createConsumer() throws IOException, InterruptedException, URISyntaxException {
-        String consumerInfo = "{\"name\":\"" + this.config.getClientId() + "\",\"format\":\"json\",\"auto.offset.reset\":\"earliest\"}";
+        Properties props = config.getProperties();
+        ObjectMapper objectMapper = new ObjectMapper();
+        String consumerInfo = objectMapper.writeValueAsString(createConsumerConfig(props));
```
ppatierno:
```suggestion
        String consumerInfo = objectMapper.writeValueAsString(createConsumerConfig(config.getProperties()));
```

java/http/java-http-consumer/src/main/java/HttpConsumer.java
```diff
@@ -84,11 +100,31 @@ public HttpConsumer(HttpConsumerConfig config, CountDownLatch messagesReceivedLa
...
+                }
+                consumerConfig.put(k, obj);
+            }
+        }
+        return consumerConfig;
+    }
+
     public void createConsumer() throws IOException, InterruptedException, URISyntaxException {
-        String consumerInfo = "{\"name\":\"" + this.config.getClientId() + "\",\"format\":\"json\",\"auto.offset.reset\":\"earliest\"}";
+        Properties props = config.getProperties();
```
ppatierno:
```suggestion
```

ppatierno (approved):
LGTM. Assuming you tried and it worked :-)

kyguy (approved):
Looks great @OwenCorrigan76!

scholzj (approved):
LGTM Assuming it works.

OwenCorrigan76:
@scholzj Yes it works as requested. Thanks

PR #135, opened 2023-08-28
Add Scram-Sha-512 to Client Examples

OwenCorrigan76:
This YAML in the Client Examples will add SCRAM-SHA-512 authentication. This will be an addition to TLS authentication which currently exists in these examples.


OwenCorrigan76:
@scholzj @kyguy @ppatierno Apologies for the multiple commits. I know it's very messy. The yaml does work though.

java/http/java-http-consumer/src/main/java/HttpConsumerConfig.java
scholzj, line 1: Why is this change here? It seems to be completely unrelated.

java/kafka/deployment-scram-sha-512.yaml
```diff
@@ -0,0 +1,57 @@
...
+            - name: KAFKA_KEY_SERIALIZER
+              value: "org.apache.kafka.common.serialization.StringSerializer"
+            - name: KAFKA_VALUE_SERIALIZER
+              value: "org.apache.kafka.common.serialization.StringSerializer"
+            - name: KAFKA_SECURITY_PROTOCOL
+              value: "SASL_SSL"
+            - name: KAFKA_SASL_MECHANISM
+              value: "SCRAM-SHA-512"
+            - name: KAFKA_SASL_JAAS_CONFIG
+              value: "org.apache.kafka.common.security.scram.ScramLoginModule required username='my-user' password='tsJ3YeyLR1wf0z8vNBLxM10pH0tA3KYq';"
```
scholzj, line 50: You create the user -> but you hardcode this information here so it would not work for users who deploy this YAML. You have to mount it from the User Secret.

java/kafka/deployment-scram-sha-512.yaml
scholzj, line 1: I would expect this to mirror https://github.com/strimzi/client-examples/blob/main/java/kafka/deployment-ssl-auth.yaml with the only difference being the authentication. I.e. create the topics, users, consumers, producers and streams API deployments.

scholzj (changes requested):

OwenCorrigan76:
Closing this PR as I've made a mess of the commit history etc. Will make the changes and fix in a separate PR.

PR #136, opened 2023-08-29
Add Scram-Sha-512

OwenCorrigan76:
This PR adds support for Scram-Sha-512 and related example deployment file.

### DIFF
--- a/java/kafka/deployment-scram-sha-512.yaml
+++ b/java/kafka/deployment-scram-sha-512.yaml
@@ -1,0 +1,204 @@
+apiVersion: kafka.strimzi.io/v1beta1
+kind: KafkaTopic
+metadata:
+  name: my-topic
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  replicas: 3
+  partitions: 12
+---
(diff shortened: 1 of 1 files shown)


OwenCorrigan76:
@scholzj Should I change Bootstrap Server to be 9093 as TLS is used? In this yaml, it is set to 9092!

java/kafka/deployment-scram-sha-512.yaml
```diff
@@ -0,0 +1,204 @@
...
+  replicas: 3
+  partitions: 12
+---
+apiVersion: kafka.strimzi.io/v1beta1
+kind: KafkaUser
+metadata:
+  name: java-kafka-producer
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
```
scholzj, line 27: Why not add the ACLs as you have it in the SSL Auth example?

OwenCorrigan76: Added now

scholzj (approved):

ppatierno (approved):

PR #149, opened 2024-08-04
Update dependencies and minor fixes and improvements

scholzj:
This PR updates the dependencies of the client examples. It also does some minor fixes / improvements:
* Updated the Dockerfiles to not use the deprecated `ENV` syntax
* Fixes the Java HTTP Consumer example to properly configure the consumer name
* Disables the Garbage collection logging that seems disrupting the log output of the running containers and does not seem to add any value to the examples


ppatierno (approved):
LGTM. Thanks!

